PR #102, opened 2022-01-27
Ci/deploy be

ktrz:


.github/workflows/be-deploy.yml
```diff
@@ -1,6 +1,8 @@
 name: Deploy backend (main)
 
 on:
+  pull_request:
+    branches: [main]
```
dustinsgoodman, line 5: 😂 that explains a lot

.github/workflows/be-deploy.yml
```diff
@@ -37,7 +39,7 @@ jobs:
         run: npm i -g serverless@2.x
 
       - name: Deploy
-        run: sls deploy --stage dev
+        run: sls deploy --config serverless-ci.yml --stage dev
```
dustinsgoodman, line 42: I don't think this is the right solution here as now we'll have to maintain 2 separate yaml files for this project which could cause permission issues and other problems down the line. I think we need to find a way to dynamically set the profile in the regular `serverless.yml`. I just remembered how I did this on a separate project!!! Here's the steps:

This project tries to use a multi-service setup which is why it's in 2 files but we can consolidate to one serverless yaml for this:
1. Create stage specific variables like I did here: https://github.com/dustinsgoodman/serverless-template/blob/main/serverless.common.yml#L19-L30
2. That variable can then be used to set the profile variable in the main `serverless.yaml` like so: https://github.com/dustinsgoodman/serverless-template/blob/main/services/public-api/serverless.yml#L20 - this can be simplified down though to just `${${self:custom.stage_env.${self:provider.stage}.PROFILE, 'dev'}`
3. Set the stage to prod above and it will use that definition instead and we can separate our local setup. 😄

dustinsgoodman: PLEASE IGNORE ALL OF THE ABOVE! Why? Because [serverless 3 just came out with stage parameters](https://www.serverless.com/blog/serverless-framework-v3-is-live) out of the box so you don't need to do this stupid hack of mine

jtomchak: Definitely want to make that `--stage` dynamic with something like 'local', 'dev', 'prod' in the example above if those env's aren't set it'll default to 'dev'. So we override it to local when working, and set it to prod for public release

jtomchak: Are we on Serverless 3?

dustinsgoodman: No, but the migration is super doable given our state.

ktrz: I think you misunderstood my intention. The idea behind a separate YAML file was to not set `profile` at all for CI so it would use the environment variables provided. But I figured out how to configure the profile using `serverless config credentials` command so we're good. I've upgraded Serverless to v3 anyway

.github/workflows/be-deploy.yml
ktrz: I don't know what do you mean. But it's removed now

dustinsgoodman: It explained why the CI wasn't running on branch PR.

starter-dev-backend/serverless.yml
```diff
@@ -13,9 +13,8 @@ custom:
 provider:
   name: aws
   profile: starterdev
-  stage: production
+  stage: dev
```
dustinsgoodman, line 16: Change this to:
```suggestion
  stage: ${sls:stage}
  region: ${aws:region}
```

See https://www.serverless.com/framework/docs/providers/aws/guide/variables#referencing-serverless-core-variables and https://www.serverless.com/framework/docs/providers/aws/guide/variables#referencing-aws-specific-variables

dustinsgoodman (changes requested):

dustinsgoodman (approved):

Issue #104, opened 2022-01-31
[Angular + NgRx + SCSS] 2 Create login page


lindakatcodes:
Create login page

PR #111, opened 2022-02-02
[Angular - NgRx - SCSS] #104 - login flow

lindakatcodes:
In this PR:

- Sets up services for allowing a user to authenticate with GitHub
- Handles routing from sign in flow to redirect to home page
- Creates an auth guard to protect the home route and redirect an unauthenticated user to the sign in page
- Adds styles to the sign in page

Fixes: #104


angular-ngrx-scss/.env.example
```diff
@@ -0,0 +1,5 @@
+JWT_SECRET=7ZkCMuD69N2WBBDCDzxTTYvsNG8qPJHWeNYe37iZTkmQLUmujZUZ3RnkqP6yfxTU
```
TapaiBalazs, line 1: I hope this is not a valid jwt secret.

- [ ] make sure that the example env file does not contain a valid or otherwise usable jwt secret

angular-ngrx-scss/src/app/auth/auth.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
TapaiBalazs, line 1: - [ ] Please delete the test file if there's nothing worth testing in the component.

angular-ngrx-scss/src/app/auth/redirect/redirect.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
TapaiBalazs, line 1: - [ ] Delete the test file if you don't intend to implement any tests

angular-ngrx-scss/src/app/auth/redirect/redirect.component.ts
```diff
@@ -0,0 +1,35 @@
...
+    private authService: AuthService,
+    private route: ActivatedRoute,
+    private router: Router,
+  ) {}
+
+  ngOnInit(): void {
+    /**
+     * When we get the code from GitHub, pass it to the authService and redirect the user to the home page
+     */
+    this.route.queryParamMap
```
TapaiBalazs, line 21: I think I would put this into a RedirectGuard where I'd return `/` or back to `/signin`.

For more on this please see: https://juristr.com/blog/2018/11/better-route-guard-redirects/

TapaiBalazs: This is not a request to change though, because putting this into a guard will not render the `Redirecting...` text. I guess we would need a spinner or a loading indicator using a component portal for that.

angular-ngrx-scss/src/app/auth/services/auth.guard.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { TestBed } from '@angular/core/testing';
```
TapaiBalazs, line 1: - [ ] please write unit tests to cover the logic in this guard.

angular-ngrx-scss/src/app/auth/services/auth.service.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { TestBed } from '@angular/core/testing';
+
+import { AuthService } from './auth.service';
```
TapaiBalazs, line 3: - [ ] Please write unit tests using the httpClientTestingModule

angular-ngrx-scss/src/app/auth/services/token.service.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { TestBed } from '@angular/core/testing';
```
TapaiBalazs, line 1: - [ ] Delete this test file please because this service is very basic, it does not need to be tested imho

angular-ngrx-scss/src/app/home/home.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
TapaiBalazs, line 1: - [ ] delete the test file if there's nothing worth testing or `xdescribe` (if Karma-Jasmine) or `describe.skip(` (if Jest) if you plan on adding future tests to it.

angular-ngrx-scss/src/app/home/home.component.ts
```diff
@@ -0,0 +1,12 @@
+// import { Component, OnInit } from '@angular/core';
```
TapaiBalazs, line 1: Why is this commented out entirely? 🤔

TapaiBalazs (changes requested):
Delete the test files where they are not needed, write tests where they are needed. Also Home component ts file got commented out entirely, I'm not sure that's intended

angular-ngrx-scss/.env.example
lindakatcodes: I'll clear it out - it's copied straight from the ang/apollo/tw project so it's there too 😬 lol but yeah should have taken that out anyways.

angular-ngrx-scss/src/app/home/home.component.ts
lindakatcodes: For now, because it was failing the linter from everything being empty. The next task is what will actually start to fill this out with data. It's a temporary thing - I need the component in place to ensure the auth routing works, but it doesn't need to actually do anything yet.

angular-ngrx-scss/src/app/home/home.component.spec.ts
lindakatcodes: left this as `xdescribe` since the functionality won't exist until later in the project. can decide then if the test file should be improved or removed.

angular-ngrx-scss/src/app/home/home.component.ts
lindakatcodes: Didn't realize this was causing other errors in the app, so fixed this to remove the parts it didn't need. It's now not commented out, and still doesn't have any functionality.

angular-ngrx-scss/src/app/auth/services/auth.service.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    const req = httpController.expectOne({
+      method: 'POST',
+      url: expectedPostUrl,
+    });
+
+    req.flush(expectedAuthResponse);
+  });
+
+  it('checks if a user is authenticated', () => {
```
lindakatcodes, line 62: @TapaiBalazs Would love your help here - the test suite doesn't always make these mock calls in order or run these in order, so the test sometimes passes and sometimes doesn't. Is there a better way I can more consistently give it the value I want to test the functionality?

angular-ngrx-scss/src/app/auth/redirect/redirect.component.ts
lindakatcodes: Fair point here. An improvement to this is potentially underway in another PR from Morgan. Will come back to this if / when those changes are approved.

angular-ngrx-scss/src/app/app.component.spec.ts
```diff
@@ -2,7 +2,7 @@ import { TestBed } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
 import { AppComponent } from './app.component';
 
-describe('AppComponent', () => {
+xdescribe('AppComponent', () => {
```
ktrz, line 5: if we skip this tests we can as well remove the file

angular-ngrx-scss/src/app/auth/auth.component.ts
```diff
@@ -0,0 +1,23 @@
...
+@Component({
+  selector: 'app-auth',
+  templateUrl: './auth.component.html',
+  styleUrls: ['./auth.component.scss'],
+})
+export class AuthComponent {
+  signInGroup = new FormGroup({
+    csrfToken: new FormControl(''),
+    callbackUrl: new FormControl(''),
+  });
```
ktrz, line 14: do we actually use this form for anything?

angular-ngrx-scss/src/app/auth/auth.component.html
```diff
@@ -0,0 +1,7 @@
+<div class="page">
+  <form class="form" [formGroup]="signInGroup" (ngSubmit)="onSubmit()">
+    <input type="hidden" class="input" formControlName="csrfToken" />
+    <input type="hidden" class="input" formControlName="callbackUrl" />
```
ktrz, line 4: are those 2 hidden inputs used for anything?

angular-ngrx-scss/src/app/auth/auth.component.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  constructor(private authService: AuthService) {}
+
+  onSubmit() {
+    this.authService.signIn().subscribe(({ redirectUrl }) => {
+      window.location.href = redirectUrl;
+    });
```
ktrz, line 21: If we want this app to showcase NgRx I think we should move the auth flow to the NgRx effects. But we can do it while redoing this solution after BE is modified

angular-ngrx-scss/src/app/auth/services/auth.guard.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  canActivate():
+    | Observable<boolean | UrlTree>
+    | Promise<boolean | UrlTree>
+    | boolean
+    | UrlTree {
+    if (this.authService.isAuthenticated()) {
+      return true;
+    }
+    return this.router.createUrlTree(['/signin']);
```
ktrz, line 20: just for reference this could be simplified to
```suggestion
    return this.authService.isAuthenticated() || this.router.createUrlTree(['/signin']);
```

But not necessary to change if you prefer the more verbose approach

angular-ngrx-scss/src/app/auth/services/auth.service.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    TestBed.configureTestingModule({
+      imports: [HttpClientTestingModule],
+      providers: [
+        {
+          provide: TokenService,
+          useValue: {
+            getToken() {
+              return token;
+            },
+            saveToken(token: string) {
```
ktrz, line 27: you can omit the param if it's not used
```suggestion
            saveToken() {
```

angular-ngrx-scss/src/app/auth/services/auth.service.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+              return;
+            },
+            removeRefreshToken() {
+              return;
+            },
+          },
+        },
+      ],
+    });
+    tokenService = TestBed.inject(TokenService);
```
ktrz, line 40: I think this line is not necessary
```suggestion
```

angular-ngrx-scss/src/app/auth/services/auth.service.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  });
+
+  it('should redirect the user to the sign in link', () => {
+    const expectedSignInResponse = {
+      redirectUrl: `${environment.apiUrl}/auth/signin`,
+    };
+
+    authService.signIn().subscribe((res) => {
+      expect(res).toEqual(expectedSignInResponse as SignInResponse);
+    });
```
ktrz, line 52: to avoid casting you should be able to explicitly type the `expectedSignInResponse`
```suggestion
    const expectedSignInResponse: SignInResponse = {
      redirectUrl: `${environment.apiUrl}/auth/signin`,
    };

    authService.signIn().subscribe((res) => {
      expect(res).toEqual(expectedSignInResponse);
    });
```

angular-ngrx-scss/src/app/auth/services/auth.service.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    const expectedAuthResponse = {
+      access_token: '12345',
+      bearer: '',
+      scope: '',
+    };
+    const expectedPostUrl = `${environment.apiUrl}/auth/signin/callback`;
+
+    authService.getToken('code').subscribe((res) => {
+      expect(res).toEqual(expectedAuthResponse as AuthResponse);
+    });
```
ktrz, line 72: same suggestion as above. In general, when you need to cast explicitly it is very likely that something was not typed properly before
```suggestion
    const expectedAuthResponse: AuthResponse = {
      access_token: '12345',
      bearer: '',
      scope: '',
    };
    const expectedPostUrl = `${environment.apiUrl}/auth/signin/callback`;

    authService.getToken('code').subscribe((res) => {
      expect(res).toEqual(expectedAuthResponse);
    });
```

ktrz (approved):
Looks good! Added a couple of suggestions but you can also address them in the follow-up PR to unblock your remaining PRs

angular-ngrx-scss/src/app/auth/services/auth.service.spec.ts
lindakatcodes: The param for `saveToken` is used to store it in local storage. :)

angular-ngrx-scss/src/app/auth/services/auth.service.spec.ts
lindakatcodes: Will double check this later.

angular-ngrx-scss/src/app/auth/services/auth.service.spec.ts
lindakatcodes: Will update this when we do the OAuth update!

angular-ngrx-scss/src/app/auth/auth.component.ts
lindakatcodes: Fair point - adding it to my list of things to update when we do the OAuth fix.

Issue #124, opened 2022-02-15
[VUE + TS + Quasar]  Create navigation header


hdJerry:
## Background

As a user, I should see a header with the GitHub logo as a home button on the left, and my user profile picture on the right attached to a dropdown menu. The menu should show links for navigating to my profile and a sign-out link.

## Acceptance

- [x] Top nav
  - [x] GitHub logo on the left
  - [x] User avatar with dropdown on the right
  - [x] Navigation to profile in the dropdown (can point to not yet existing page)
  - [x] Logout link in the dropdown (can point to not yet existing functionality yet)
- [x] Placeholder for main content

## Notes

PR #151, opened 2022-02-18
Component/navbar (#124)

AllanJeremy:
Added the `NavHeader` component as well as its related components (`Logo` and `UserAvatar`).

- Added stories for each of the components
- Added unit tests for each of the components
- Added `@pinia/testing` for mock stores when doing unit testing

and probably some other minor details

### DIFF
--- a/vue3-ts-quasar/src/components/Logo/Logo.stories.ts
+++ b/vue3-ts-quasar/src/components/Logo/Logo.stories.ts
@@ -3,7 +3,14 @@ import Logo from './Logo.vue';
 export default {
   title: 'component/Logo',
   component: Logo,
-  argTypes: {},
+  argTypes: {
+    dark: { type: 'boolean' },
+    size: { type: 'number' },
+    darkImg: { type: 'string' },
+    darkLgImg: { type: 'string' },
+    lightLgImg: { type: 'string' },
+    lightImg: { type: 'string' },
+  },
 };
 
 const Template = (args) => ({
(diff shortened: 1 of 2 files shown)


hdJerry:
## Stroy Book

For this args to reflect you will need to add each prop in the argTypes.
eg.
```
dark: {}
```

## Worth Nothing
for example, you got a component name as NavBar instead of `<NavBar />` your could do `<nav-bar />`. It makes it steal looks like an HTML tag. You will notice how quasar did theirs too using `<q-bla-bla><q-bla-bla />`

AllanJeremy:
Please clarify on the `dark: {}` part. What component? What does not reflect?
PS, the default `dark` attribute currently works but you can't see the image because it is white against a dark background. Would changing the background when it is dark in storybook help?

## On kebab-case
Got it, thanks. Usually have it as `PascalCase` to distinguish between `CustomComponents` and `framework-components` in an instant.

hdJerry:
> [Vue3 + TS + Quasar] component: Added logo component stories

The logo component story compared to how you did the avatar component story.
Check both argTypes.

AllanJeremy:
Mind if I refactor the `kebab-case` to `PascalCase`?

hdJerry (approved):

hdJerry:
> Mind if I refactor the `kebab-case` to `PascalCase`?

No need big man

AllanJeremy:
> > Mind if I refactor the `kebab-case` to `PascalCase`?
> 
> No need big man

^ This will be an opensource project, so there probably is one 😅

PR #161, opened 2022-02-21
[Angular - NgRx - SCSS] - Get profile info from API

SDaian:


lindakatcodes (changes requested):
I notice a few things we're missing here:

- The "bio" field is missing entirely, and is part of the requirements (called "description" in the ticket). That will need to be added to the `UserState` and anywhere else we update that state.
- I think we should have "stars" in all the places where UserState is, and it's currently only showing in the UserState and reducer files.
- Also for the stars - I know we don't get that info directly from the REST API, so we should be doing a bit of calculation to get that data. It looks like the users API call returns a `starred_url` field that I think will provide the other call we'll need to make to get that count.

Other than those, I think this looks good! 👍

lindakatcodes:
This ended up being included in a different PR for the profile page. Since it's duplicating the same changes and we need the other one, choosing to close this one. Appreciate the effort, Daian!

PR #179, opened 2022-03-02
Vue/component/fix props names tab header repo card

hdJerry:
- Typo in props in TabHeader . respositories -> repositories
- Prop names should be kebab-case rather than TitleCase in ProfilePageLayout
- Last updated in RepoCard should be a date (that ends up formatted - I'll do the formatting)


AllanJeremy (changes requested):
Last updated should in its end state be a `Date` object, this does serve as a good starting point though.

AllanJeremy (approved):
Looks good. Will need to make adjustments to the default `lastUpdated` value when fetching data but other than that, looks good.

PR #181, opened 2022-03-04
[Angular - NgRx - SCSS] #118: Implement User Profile Page

### LINKED ISSUE #118: [Angular - NgRx - SCSS] Create profile page layout
# Create profile page layout

## Background

As a user, I can navigate to my profile page and see my profile information, as well as a list of all my repositories. I can also see options to search my repositories, sort them, and filter them. (Functionality does not need to work yet for these options, they just need to be visible.)

## Acceptance

- [ ] 2 column layout
- [ ] sub-header spanning both columns, showing book icon and repositories title - should be sticky on scroll
- [ ] Left side
    - [ ] user profile card
- [ ] Right side - 
    - [ ] search / filter / sort component
    - [ ] card for all user repositories

## Notes

Reshurum:
Resolves #118 

~~**NOTE:** Branch is behind and doesn't include changes from auth refactor. Merge conflicts are likely to occur.~~


angular-ngrx-scss/src/app/home/profile/profile-about/profile-about.component.html
```diff
@@ -0,0 +1,79 @@
...
+          profile?.user?.email
+        }}</a>
+      </div>
+      <div *ngIf="profile?.user?.blog">
+        <span class="icon" appOcticon="link"></span>
+        <a [href]="profile?.user?.blog">{{ profile?.user?.blog }}</a>
+      </div>
+      <div *ngIf="profile?.user?.twitter_username">
+        <span class="icon twitter-icon">
+          <svg
```
lindakatcodes, line 49: This seems fairly small, but if the octicons package doesn't have a Twitter icon, would like to have this as a file in the assets folder that we pull in here instead of the direct svg, unless it's massively more complex to do that way.

angular-ngrx-scss/src/app/home/profile/profile-about/profile-about.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
lindakatcodes, line 1: Can be deleted

angular-ngrx-scss/src/app/home/profile/profile-nav/profile-nav.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
lindakatcodes, line 1: Can be deleted

angular-ngrx-scss/src/app/home/profile/profile-repos/profile-repos.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
lindakatcodes, line 1: Would potentially be good to update this with a small test on the time functionality you did? Otherwise, can be deleted.

angular-ngrx-scss/src/app/home/profile/profile.component.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
lindakatcodes, line 1: Can be deleted

angular-ngrx-scss/src/app/state/profile/profile.actions.spec.ts
```diff
@@ -0,0 +1,9 @@
+import * as fromProfile from './profile.actions';
```
lindakatcodes, line 1: Can be deleted

angular-ngrx-scss/src/app/state/profile/profile.effects.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { TestBed } from '@angular/core/testing';
```
lindakatcodes, line 1: This would be the best file to keep and add actual tests to, to double check our side effects.

angular-ngrx-scss/src/app/state/profile/profile.reducer.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { reducer, initialState } from './profile.reducer';
```
lindakatcodes, line 1: Can be deleted

angular-ngrx-scss/src/app/state/profile/profile.selectors.spec.ts
```diff
@@ -0,0 +1,3 @@
+describe('Profile Selectors', () => {
```
lindakatcodes, line 1: Can be deleted

angular-ngrx-scss/src/app/state/user/user.effects.spec.ts
```diff
@@ -2,18 +2,19 @@ import { TestBed } from '@angular/core/testing';
 import { provideMockActions } from '@ngrx/effects/testing';
 import { Action } from '@ngrx/store';
 import { Observable, of } from 'rxjs';
+import { fetchUserData, fetchUserDataSuccess } from '.';
```
lindakatcodes, line 5: This should still have `/user.actions`, to keep things clear where this comes from

lindakatcodes (commented):
A few general questions/things that don't belong to a specific section:
- I think the `yarn.lock` needs to be removed. For this particular showcase it's using npm, so we'll want to only keep the `package-lock` file.
- You know this, but will need to be rebased with the main branch to pull in the auth changes and fix the conflicts. 
- Will also need to see why the test check is failing - could be simply because of merge conflicts, but should verify this. I'm also leaving some comments on test files I don't think we need - the intent for these is to show best practices, so some files that don't have a lot to test we should remove the default tests. A few of the files should have actual tests written for them, and it would be nice to do them in this PR but we might also want to improve on them later - open to discussion on that point.

Did not have enough time to fully review everything, but will add this review for now and do another pass on the profile and state sections next week! 👍

angular-ngrx-scss/src/app/user/services/user.service.ts
```diff
@@ -10,12 +16,93 @@ import { environment } from 'src/environments/environment';
...
+  getAuthenticatedUserInfo(): Observable<UserState> {
+    const url = `${environment.githubUrl}/user`;
+
+    return this.http.get<UserApiResponse>(url).pipe(
+      map((data) => ({
+        avatar: data.avatar_url,
+        bio: data.bio,
+        blog: data.blog,
+        company: data.company,
+        email: data.email,
```
lindakatcodes, line 28: Can remove email.

angular-ngrx-scss/src/app/user/services/user.service.ts
```diff
@@ -10,12 +16,93 @@ import { environment } from 'src/environments/environment';
...
+      username,
+    )}`;
+
+    return this.http.get<UserApiResponse>(url).pipe(
       map((data) => ({
         avatar: data.avatar_url,
+        bio: data.bio,
+        blog: data.blog,
+        company: data.company,
+        email: data.email,
```
lindakatcodes, line 50: Can remove email

angular-ngrx-scss/src/app/user/services/user.service.ts
```diff
@@ -10,12 +16,93 @@ import { environment } from 'src/environments/environment';
...
+        location: data.location,
+        name: data.name,
+        twitter_username: data.twitter_username,
         username: data.login,
       })),
     );
   }
+
+  getUserOrganizations(username: string): Observable<UserOrgsState[]> {
+    const url = `https://api.github.com/users/${encodeURIComponent(
```
lindakatcodes, line 62: This should use the environment url variable for consistency.

angular-ngrx-scss/src/app/user/services/user.service.ts
```diff
@@ -10,12 +16,93 @@ import { environment } from 'src/environments/environment';
...
+          id: org.id,
+          login: org.login,
+          avatar_url: org.avatar_url,
+        })),
+      ),
+    );
+  }
+
+  getUserRepos(username: string): Observable<UserReposState[]> {
+    const url = `https://api.github.com/users/${encodeURIComponent(
```
lindakatcodes, line 78: This should use the environment url variable for consistency.

angular-ngrx-scss/src/app/user/services/user.service.spec.ts
```diff
@@ -20,17 +20,35 @@ describe('UserService', () => {
...
+      location: '',
       login: 'lindakatcodes',
+      name: '',
+      twitter_username: '',
     };
 
     httpClientSpy.get.and.returnValue(of(expectedHttpResponse).pipe(delay(0)));
 
-    userService.getUserInfo().subscribe((res) => {
+    userService.getAuthenticatedUserInfo().subscribe((res) => {
```
lindakatcodes, line 51: Potentially needs changing back if we don't keep both calls.

angular-ngrx-scss/src/app/state/user/user.state.ts
```diff
@@ -1,5 +1,14 @@
 export interface UserState {
   avatar: string;
+  bio: string;
+  blog: string;
+  company: string;
+  email: string;
```
lindakatcodes, line 6: Can remove email

angular-ngrx-scss/src/app/state/user/user.reducer.ts
```diff
@@ -4,6 +4,15 @@ import { fetchUserDataSuccess } from './user.actions';
 
 const initialUserState: UserState = {
   avatar: '',
+  bio: '',
+  blog: '',
+  company: '',
+  email: '',
```
lindakatcodes, line 10: Can remove email

angular-ngrx-scss/src/app/state/user/user.effects.ts
```diff
@@ -15,7 +15,7 @@ export class UserEffects {
     return this.actions$.pipe(
       ofType(fetchUserData),
       switchMap(() =>
-        this.userService.getUserInfo().pipe(
+        this.userService.getAuthenticatedUserInfo().pipe(
```
lindakatcodes, line 18: Potentially needs to be switched back if we don't keep both calls.

angular-ngrx-scss/src/app/state/user/user.effects.spec.ts
```diff
@@ -2,18 +2,19 @@ import { TestBed } from '@angular/core/testing';
...
 
 describe('UserEffects', () => {
   let actions$: Observable<Action>;
   let effects: UserEffects;
   let userServiceMock: any;
+
   beforeEach(() => {
     userServiceMock = jasmine.createSpyObj('UserService', {
-      getUserInfo: () => {
+      getAuthenticatedUserInfo: () => {
```
lindakatcodes, line 17: This and line 56 can potentially be changed back if we don't keep both calls.

angular-ngrx-scss/src/app/state/profile/profile.state.ts
```diff
@@ -0,0 +1,164 @@
+import { UserState } from '../user';
+
+export interface ProfileState {
+  user?: UserState;
```
lindakatcodes, line 4: I don't think user should be optional here - the only user we track is the one that's signed in, and they can't view any of this if they don't exist.

angular-ngrx-scss/src/app/state/profile/profile.selectors.ts
```diff
@@ -0,0 +1,11 @@
...
+import { ProfileState } from './profile.state';
+
+export const profileFeatureKey = 'profile';
+export const selectProfileState =
+  createFeatureSelector<ProfileState>(profileFeatureKey);
+
+export const selectProfile = createSelector(
+  selectProfileState,
+  (state: ProfileState) => state,
+);
```
lindakatcodes, line 11: Mentioned in another comment, but I'd add a selector here for the repos.

angular-ngrx-scss/src/app/state/profile/profile.reducer.ts
```diff
@@ -0,0 +1,18 @@
+import { Action, createReducer, on } from '@ngrx/store';
+import { fetchProfileSuccess } from './profile.actions';
+import { ProfileState } from './profile.state';
+
+export const initialState: ProfileState = {};
```
lindakatcodes, line 5: I feel like this should have the default values for the values we expect here, instead of being an empty object. But perhaps it's not needed? @ktrz @morgnism would appreciate your thoughts here, not sure if there's a preferred best practice for that or not.

lindakatcodes (commented):
Overall comments:

- I think the profile folder (with all the component pieces) should live outside of home. Each overall page should be in the same level of the project - so home is the main dashboard view, profile is its own page, single repo views will be their own page. So I think that entire profile folder should be moved up a level to be equal to home.

The changes you have here in the user state will overlap with what Daian has in his open PR. There's a few small naming differences, but otherwise it'll be similar data. I see two paths forward:
- Option 1: We finalize and merge Daian's PR, and then rebase yours after that's merged in and adjust whatever's needed as far as the names of fields go. There's a small change or two that his PR needs, but I can handle those if he's not going to have any hours the rest of this month.
- Option 2: We decline Daian's PR and use your version instead. 

@ktrz @dustinsgoodman Would really appreciate your input on this decision - Daian's work was done first and the ticket for Walter's PR was only meant to handle the style changes, but both solutions are valid. I feel like it might be easiest to go with option 2, but I don't want to downplay anyone's work. Thoughts?

Reshurum:
> I think the yarn.lock needs to be removed. For this particular showcase it's using npm, so we'll want to only keep the package-lock file.

That's my bad. I saw the backend was using yarn and so I thought it was standardized across the different repos. I thought we were planning on using yarn workspaces since I heard it was mentioned a couple of times in the weekly standups.

> You know this, but will need to be rebased with the main branch to pull in the auth changes and fix the conflicts.

Will do once all other comments are addressed 👍 

> Will also need to see why the test check is failing - could be simply because of merge conflicts, but should verify this. I'm also leaving some comments on test files I don't think we need - the intent for these is to show best practices, so some files that don't have a lot to test we should remove the default tests. A few of the files should have actual tests written for them, and it would be nice to do them in this PR but we might also want to improve on them later - open to discussion on that point.

I'll check this after the rebase.

angular-ngrx-scss/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 /* To learn more about this file see: https://angular.io/config/tsconfig. */
 {
+  "include": ["src/**/*.ts", "typings.d.ts"],
```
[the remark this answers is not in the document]
Reshurum: It does not. Seems like I added it in an extra spot while debugging and forgot to remove it.

angular-ngrx-scss/scripts/generate-language-colors.js
```diff
@@ -0,0 +1,28 @@
...
+
+  const colors = Object.keys(languages).reduce((prev, cur) => {
+    const color = languages[cur].color;
+    if (color) {
+      prev[cur] = color;
+    }
+    return prev;
+  }, {});
+
+  const out = new URL('../src/assets/language-colors.json', import.meta.url);
```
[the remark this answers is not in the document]
Reshurum: It's used for finding paths relative to a file. It's so the script can be run from any directory and still work.

angular-ngrx-scss/src/app/home/home-routing.module.ts
```diff
@@ -2,13 +2,19 @@ import { NgModule } from '@angular/core';
...
+import { ProfileComponent } from './profile/profile.component';
 
 const routes: Routes = [
   {
     path: '',
     component: HomeComponent,
     canActivate: [AuthGuard],
   },
+  {
+    path: ':username',
```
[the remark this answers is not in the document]
Reshurum: > Does this need a / in front of it, or does that get handled automatically?

A leading `/` isn't required for these routes.

> I wonder if it might make more sense to have this be loaded from the app-routing file instead? Because the profile page is a fully separate page from the home component, which will have the gists and top repos. Room for discussion and opinions on routing strategy here - the other Angular project almost feels a bit too complex, but this doesn't quite feel right here to me either. 🤔

We can probably simplify the routing here. I had originally done that but it made too many changes elsewhere and I wanted to avoid merge conflicts since that would be outside the scope of my patch.

angular-ngrx-scss/src/app/state/profile/profile.effects.spec.ts
Reshurum: So do I need to increase coverage here and check that HTTP requests are being fired?

angular-ngrx-scss/src/app/state/user/user.effects.spec.ts
Reshurum: Whoops. That's what happens when I don't check IDE auto-imports, haha.

angular-ngrx-scss/src/app/home/profile/profile-about/profile-about.component.html
```diff
@@ -0,0 +1,79 @@
...
+        <span class="icon" appOcticon="organization"></span>
+        <span>{{ profile?.user?.company }}</span>
+      </div>
+      <div *ngIf="profile?.user?.location">
+        <span class="icon" appOcticon="location"></span
+        ><span>{{ profile?.user?.location }}</span>
+      </div>
+      <div *ngIf="profile?.user?.email">
+        <span class="icon" appOcticon="mail"></span>
+        <a href="mailto:{{ profile?.user?.email }}">{{
```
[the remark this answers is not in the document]
Reshurum: Is there reason why they do not? I was making it work like upstream GitHub.

dustinsgoodman:
> Overall comments:
> 
> * I think the profile folder (with all the component pieces) should live outside of home. Each overall page should be in the same level of the project - so home is the main dashboard view, profile is its own page, single repo views will be their own page. So I think that entire profile folder should be moved up a level to be equal to home.
> 
> The changes you have here in the user state will overlap with what Daian has in his open PR. There's a few small naming differences, but otherwise it'll be similar data. I see two paths forward:
> 
> * Option 1: We finalize and merge Daian's PR, and then rebase yours after that's merged in and adjust whatever's needed as far as the names of fields go. There's a small change or two that his PR needs, but I can handle those if he's not going to have any hours the rest of this month.
> * Option 2: We decline Daian's PR and use your version instead.
> 
> @ktrz @dustinsgoodman Would really appreciate your input on this decision - Daian's work was done first and the ticket for Walter's PR was only meant to handle the style changes, but both solutions are valid. I feel like it might be easiest to go with option 2, but I don't want to downplay anyone's work. Thoughts?

@lindakatcodes can you give me more context as to what we lose by option 2? I think you're saying both options yield the exact same result but I'm wondering if there's something missing that I'm not following.  Leaning towards option 2 because it's faster but I'd like to better understand the specifics.

angular-ngrx-scss/src/app/home/profile/profile-nav/profile-nav.component.html
```diff
@@ -0,0 +1,10 @@
+<nav>
+  <ul>
+    <li class="active-tab">
+      <a routerLink="/{{ (profile$ | async)?.user?.username }}">
```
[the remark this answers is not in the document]
Reshurum: This is for the repositories for the user you are viewing so that's why the username is passed there. It's following GitHub's path conventions.

angular-ngrx-scss/src/app/home/profile/profile-nav/profile-nav.component.ts
```diff
@@ -0,0 +1,13 @@
...
+import { ProfileState } from 'src/app/state/profile/profile.state';
+
+@Component({
+  selector: 'app-profile-nav',
+  templateUrl: './profile-nav.component.html',
+  styleUrls: ['./profile-nav.component.scss'],
+})
+export class ProfileNavComponent {
+  @Input()
+  profile$?: Observable<ProfileState>;
```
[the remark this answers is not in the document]
Reshurum: If my last comment is correct then username will be needed. We should discuss this.

angular-ngrx-scss/src/app/home/profile/profile-repos/profile-repos.component.ts
```diff
@@ -0,0 +1,63 @@
+import { Component, Input } from '@angular/core';
+import { Observable } from 'rxjs';
+import { UserReposState } from 'src/app/state/profile/profile.state';
+import languageColors from 'src/assets/language-colors.json';
+
+const MINUTE = 60,
```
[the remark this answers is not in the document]
Reshurum: Yeah! I think that's a good idea 👍

angular-ngrx-scss/src/app/state/user/user.effects.ts
Reshurum: We should discuss this later. I think I have some missing context.

angular-ngrx-scss/src/app/state/user/user.effects.spec.ts
Reshurum: We should discuss this later. I think I have some missing context.

angular-ngrx-scss/src/app/state/profile/profile.state.ts
Reshurum: I have it like this since the initial state in the reduce will have to fill in an empty user state. I don't know what the best practice is here, meaning is it better to start with an object with empty strings and zeros or just use null until it's defined.

angular-ngrx-scss/src/app/state/profile/profile.reducer.ts
Reshurum: I don't know what the best practice here. I made the values nullable so this spot would not need to be refactored to add empty values if we added new fields.

angular-ngrx-scss/src/app/state/profile/profile.selectors.ts
Reshurum: Would these selectors be used by this PR at all?

angular-ngrx-scss/src/app/home/profile/profile.component.html
```diff
@@ -0,0 +1,30 @@
...
+        <div class="col-span-3"></div>
+        <app-profile-nav
+          class="col-span-9"
+          [profile$]="profile$"
+        ></app-profile-nav>
+      </div>
+    </div>
+
+    <div class="grid grid-cols-12 profile-body container">
+      <div class="subpage col-span-3">
```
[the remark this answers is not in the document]
Reshurum: Good catch!

angular-ngrx-scss/src/app/home/profile/profile.component.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.store.dispatch(fetchUserData());
+
+    this.route.params
+      .pipe(takeUntil(this.destroy$))
+      .subscribe(({ username }) => {
+        this.store.dispatch(fetchProfile({ username }));
+      });
+
+    this.profile$ = this.store.select(selectProfileState);
+    this.repos$ = this.profile$.pipe(map((profile) => profile.repos ?? []));
```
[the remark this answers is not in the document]
Reshurum: Yeah that makes sense to me I guess. In NgRx is it a good practice to use selectors over inputs in cases like this?

angular-ngrx-scss/src/app/user/services/user.service.ts
```diff
@@ -10,12 +16,93 @@ import { environment } from 'src/environments/environment';
 export class UserService {
   constructor(private http: HttpClient) {}
 
-  getUserInfo(): Observable<UserState> {
-    return this.http.get<UserApiResponse>(`${environment.githubUrl}/user`).pipe(
+  getAuthenticatedUserInfo(): Observable<UserState> {
```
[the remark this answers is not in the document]
Reshurum: The profile page can display information for any user on GitHub which is why there is an additional call. The nav bar at the top always has the authenticated user, and the profile page has the user details of the user specified in the profile page route.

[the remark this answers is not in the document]
Reshurum: Now maybe there can be an optimization in the case of the profile page user being the same as the authenticated user? Right now it does two calls even if it's for the authenticated user, but I think it should still work in the viewing another user that's not you case.

angular-ngrx-scss/src/app/state/profile/profile.effects.spec.ts
lindakatcodes: Doesn't have to be complete coverage (and I think the getUserInfo call is already handled in it's own test file), but something to show what a test would look like I think is what we'd want. Maybe one to make sure the calls are firing like we expect and that we get some mocked results back? (Can also make a backlog ticket for this, I don't think this is super pressing)

angular-ngrx-scss/src/app/home/profile/profile-about/profile-about.component.html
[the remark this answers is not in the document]
lindakatcodes: (Putting here for visibility) We've decided to go ahead and leave the email in, and we'll make tickets to update the other showcases. 👍

angular-ngrx-scss/src/app/home/profile/profile-nav/profile-nav.component.html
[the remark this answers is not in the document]
lindakatcodes: So I think for now, we can leave this and look again when the profile folder is moved up a level. That might still work as expected and I do like keeping to GitHub's arrangement as much as possible, but since for our showcases the profile view is the only way to view a user's repos, it may not need the username. Can revisit this though, I could be wrong!

angular-ngrx-scss/src/app/home/profile/profile-nav/profile-nav.component.ts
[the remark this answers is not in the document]
lindakatcodes: Yeah let's get the folder moved up a level and then see what we need. I think if we do keep the username (which is potentially fine) we should adjust this to pass just the username as the `Input`, instead of the full profile (since username is all we're using from the profile here).

angular-ngrx-scss/src/app/state/profile/profile.state.ts
lindakatcodes: Oh good question - @ktrz any thoughts on best practice here?

angular-ngrx-scss/src/styles.scss
```diff
@@ -1,43 +1,25 @@
 html,
 body {
```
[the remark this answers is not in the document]
Reshurum: These attributes were adding a horizontal scrollbar to the profile page when the vertical scrollbar appeared. I think that happens because 100vw doesn't subtract the width of the scrollbar in some situations? I'll check the sign-in page and see if there's a way we can get both pages styled correctly.

angular-ngrx-scss/src/app/home/profile/profile-about/profile-about.component.html
Reshurum: I embedded to avoid an extra network call for the icon. I wonder if we can get the best of both worlds somehow?

dustinsgoodman:
> > Overall comments:
> > 
> > * I think the profile folder (with all the component pieces) should live outside of home. Each overall page should be in the same level of the project - so home is the main dashboard view, profile is its own page, single repo views will be their own page. So I think that entire profile folder should be moved up a level to be equal to home.
> > 
> > The changes you have here in the user state will overlap with what Daian has in his open PR. There's a few small naming differences, but otherwise it'll be similar data. I see two paths forward:
> > 
> > * Option 1: We finalize and merge Daian's PR, and then rebase yours after that's merged in and adjust whatever's needed as far as the names of fields go. There's a small change or two that his PR needs, but I can handle those if he's not going to have any hours the rest of this month.
> > * Option 2: We decline Daian's PR and use your version instead.
> > 
> > @ktrz @dustinsgoodman Would really appreciate your input on this decision - Daian's work was done first and the ticket for Walter's PR was only meant to handle the style changes, but both solutions are valid. I feel like it might be easiest to go with option 2, but I don't want to downplay anyone's work. Thoughts?
> 
> @lindakatcodes can you give me more context as to what we lose by option 2? I think you're saying both options yield the exact same result but I'm wondering if there's something missing that I'm not following. Leaning towards option 2 because it's faster but I'd like to better understand the specifics.

Documenting the solution from a conversation from slack: we're going with option 2 as we're not losing any features in using this PR instead.

PR #207, opened 2022-03-10
feat: adds user gists

WillHutt:
image


WillHutt:
@AllanJeremy Do you have any incite to this issue I've run into?

AllanJeremy:
> @AllanJeremy Do you have any incite to this issue I've run into?

Just got to the computer, let me check it out

AllanJeremy (changes requested):
One reason you are not seeing the data is because when we access `ref`s in components/views, we do not need to use `.value`. That means that the check for `gistLoading.value` will always return `false` since the value in the DOM is actually `undefined`.

Made a suggested change~

vue3-ts-quasar/src/composables/github/types/index.ts
```diff
@@ -1,4 +1,4 @@
 //? This file simply acts as an aggregator for all types
-import Gist from './Gist';
+import GistItem from './Gist';
```
AllanJeremy, line 2: We can shorten this to 

```ts
export { default as GistItem} from './Gist'
```

vue3-ts-quasar/src/components/GistsPanel/GistsPanel.vue
```diff
@@ -11,10 +11,10 @@
       />
     </q-card-section>
     <q-separator></q-separator>
-    <q-card-section>
+    <q-card-section v-if="!gistLoading.value">
```
AllanJeremy, line 14:
```vue
<q-card-section v-if="!gistLoading">
   <!-- -->
</q-card-section>
```

WillHutt:
I resolved your comments, but that wasn't the issue. 

The issue stems from the `getUserGists` that has a watch to see when it updates the data. For some reason, the data won't update in the actual component.

vue3-ts-quasar/src/composables/github/useGists.ts
```diff
@@ -0,0 +1,21 @@
+import { Ref } from 'vue';
+import { parseQuery } from '@/components/GistsPanel/parseQuery';
```
AllanJeremy, line 2: Structurally, would recommend that we don't import dependencies from within components into composables.

Composables are meant to be reusable pieces of functionality within the application and thus adding them here would be coupling the composable to a GistsPanel dependency. This means if we wanted to use the get gists functionality with a different component/page, we'd need to pull the `parseQuery` functionality from the `GistsPanel` component.

## Suggestion
Migrate the `parseQuery` functionality into useGists. That way, it is self-contained.

vue3-ts-quasar/src/composables/github/useGists.ts
```diff
@@ -0,0 +1,21 @@
...
+
+interface UseGists {
+  getUserGists: () => {
+    data?: GistItem;
+    loading: Ref<boolean>;
+  };
+}
+
+export const useGists = (): UseGists => {
+  const getUserGists = (): any => {
```
AllanJeremy, line 15: Prefer strongly typing whenever possible. I think we should avoid the use of `any` whenever possible.

vue3-ts-quasar/src/components/GistsPanel/parseQuery.ts
```diff
@@ -0,0 +1,35 @@
...
+interface UserGistsQuery {
+  viewer: { gists: { nodes: any[] } };
+}
+
+type GistParse = {
+  acc: GistItem[];
+  gists: any;
+};
+
+export function parseQuery(data: UserGistsQuery): GistParse {
```
AllanJeremy, line 12: The return type for this is actually `GistItem[]`

AllanJeremy:
```ts
function parseQuery(data: UserGistsQuery): GistItem[]
```
This will provide you with better type inference in the `GistsPanel` component

vue3-ts-quasar/src/components/GistsPanel/parseQuery.ts
```diff
@@ -0,0 +1,35 @@
+import type { GistItem } from '../../composables/github/types';
+
+interface UserGistsQuery {
+  viewer: { gists: { nodes: any[] } };
+}
+
+type GistParse = {
```
AllanJeremy, line 7: With the changes below, `GistParse` type is no longer needed.

vue3-ts-quasar/src/composables/github/types/index.ts
```diff
@@ -1,4 +1,2 @@
 //? This file simply acts as an aggregator for all types
-import Gist from './Gist';
-
-export { Gist };
+export { default as GistItem } from './Gist';
```
AllanJeremy, line 2: Prefer named exports in types since that means we can have multiple related types in one file.
```ts
export * from './Gist';
```

vue3-ts-quasar/src/composables/github/types/Gist.ts
```diff
@@ -0,0 +1,6 @@
+export default interface GistItem {
+  id: string;
+  description?: string | null;
+  name: string;
+  url: string;
+}
```
AllanJeremy, line 6: Prefer named exports in types since that means we can have multiple related types in one file.
```ts
export interface GistItem
```

vue3-ts-quasar/src/composables/github/useGists.ts
```diff
@@ -0,0 +1,21 @@
+import { Ref } from 'vue';
+import { parseQuery } from '@/components/GistsPanel/parseQuery';
+import { useQuery, useResult } from '@vue/apollo-composable';
+import { USER_GISTS_QUERY } from './queries';
+import { GistItem } from '../github/types';
+
+interface UseGists {
```
AllanJeremy, line 7: We can move all individual related types to `src/composables/github/types/Gist.ts` and export them as interfaces/types from the same. 

Not sure if we should move the query type to `src/composables/github/types/Gist.ts`  or into `src/composables/github/queries/user-gists.query.ts` . What do you think @WillHutt ?

vue3-ts-quasar/src/composables/github/useGists.ts
```diff
@@ -0,0 +1,21 @@
...
+  getUserGists: () => {
+    data?: GistItem;
```
AllanJeremy, line 9: Please fix types:
```ts
data: Readonly<Ref<Readonly<[] | GistItem[]>>>
```

vue3-ts-quasar/src/composables/github/useGists.ts
AllanJeremy, line 15: Prefer strongly typing whenever possible. I believe we should avoid the use of `any` whenever possible.

Remove the `any` once the return types in the `UseGists` interface have been correctly set.

AllanJeremy (changes requested):
Hey, great job so far! Functionally, it works perfect. 
However, got a few structural and type related issues that would need resolved.

Feel free to ping me incase of anything @WillHutt .

Cheers.

vue3-ts-quasar/src/composables/github/useGists.ts
```diff
@@ -0,0 +1,37 @@
...
+                    url: gist.url,
+                  },
+                ]
+              : acc,
+          [],
+        );
+        return [...acc, ...gists];
+      }, []),
+    );
+    console.log(gists);
```
hdJerry, line 33: You can remove the console log

PR #238, opened 2022-04-29
Add Currently Authenticated User's Gists to Sidebar

### LINKED ISSUE #230: [CRA-RXJS-SC] Get Gists from GitHub API
# Get Gists from GitHub API

## Background

We want to show the user a list of their gists.

## Acceptance

- [ ] Get gists from GitHub API
- [ ] For display on main page; data needed:
    - [ ] Name of gist
    - [ ] html_url

## Notes

jhmsly:
Resolves #230

### DIFF
--- a/cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
+++ b/cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
@@ -2,20 +2,16 @@ import { useEffect, useState } from 'react';
 import { GITHUB_URL_BASE, GISTS_URL } from '../../constants/url.constants';
 import { Gist, GistWithFilename } from '../../interfaces/gists.interfaces';
 import { fromFetchWithAuth } from '../auth/from-fetch-with-auth';
-import {
-  filter,
-  map,
-  Subscription,
-  Observable,
-  switchMap,
-  tap,
-  zip,
-} from 'rxjs';
+import { filter, map, Subscription, tap } from 'rxjs';
 
 export function useGists(): GistWithFilename[] {
   const [state, setState] = useState<GistWithFilename[]>([]);
   const [currentUser, setCurrentUser] = useState<string>('');
 
+  /**
+   * @todo Refactor getting the currently authenticated user's username into
+   *       a global context.
+   */
   const getCurrentUser = () =>
     fromFetchWithAuth<{
       login: string;
--- a/cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
+++ b/cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
@@ -40,14 +40,21 @@ export function useGists(): GistWithFilename[] {
     )
       .pipe(
         filter((gists) => !!gists.length),
-        switchMap((gists: Gist[]) => {
-          const requests = gists.map(createGistRequest);
+        map((gists) => {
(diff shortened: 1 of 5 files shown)


cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
```diff
@@ -0,0 +1,87 @@
...
+    }>(`${GITHUB_URL_BASE}/user`, {
+      selector: (response: Response) => {
+        return response.json();
+      },
+    });
+
+  useEffect(() => {
+    getCurrentUser().subscribe((user) => {
+      setCurrentUser(user.login);
+    });
```
danecando, line 31: Is there not some global state for the logged in user available and just pass it in as a prop? I don't think we want to refetch the user in a bunch of different places

cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
```diff
@@ -0,0 +1,87 @@
...
+  return state;
+}
+
+function createGistRequest(gist: Gist): Observable<Gist> {
+  return fromFetchWithAuth<Gist>(gist.url, {
+    selector: (response: Response) => {
+      return response.json();
+    },
+  });
+}
```
danecando, line 70: Is there not a way to just fetch a list of the most recent gists with the data needed on them? It's been a while since I did one of these but I don't remember having to do it like this. Seems like too much work going on here just to get and parse a small list of items

danecando (commented):
Just a couple of comments. RxJS code generally looks overly complicated and icky to me but probably just out of ignorance but I do think we should be able to get what we need in 1 request (aside from the user context)

cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
jhmsly: There isn't but I can look into making one. This would just move this function elsewhere.

cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
jhmsly: I'm not sure what you mean with this? This is what I'm doing with this function.

danecando: It looks like it makes a request to get a list of gists, and then creates individual requests to fetch each gist based on that

cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
dustinsgoodman: @ktrz was talking to Ben Lesh about this. I don't know if they came up with a solution though.

dustinsgoodman:
Just reviewed the design of this - please add more vertical padding between the gist list items. Here's an example from the next version:

https://next-react-query-tailwind.starter.dev/
image

danecando (approved):
LGTM

dustinsgoodman:
> The code on this branch has been updated, correcting the excessive queries for user gists and combining it all into a single request. It has been noted in `use-gists.ts` that fetching the currently authenticated user's name should be refactored into a global context. This remaining issue should be moved to a separate ticket.

Filed as https://github.com/thisdot/starter.dev-github-showcases/issues/281

PR #246, opened 2022-05-04
feat: adds userdropdown component

WillHutt:
# Overview
- adds navbar component 
- adds logout functionality

# Screenshots

## Login page
image

## Main page
image

## Main page w/navbar opened
image

## Profile page w/navbar
image

## A repo page w/navbar
image


remix/app/components/Navbar/NavBar.tsx
```diff
@@ -0,0 +1,34 @@
+import * as styles from './NavBar.classNames';
+import UserDropdownView from '../UserDropdown/UserDropdown.view';
+import { Link } from 'remix';
+import { GitHubLogo } from '../Icons/GitHubLogo';
+
+type navBarProps = {
```
danecando, line 6: Type names should start with a capital ie `NavBarProps`

remix/app/components/UserDropdown/UserDropdown.view.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <div className="py-1">
+            {username && (
+              <Menu.Item>
+                <Link to={`/${username}`} className={styles.menuBtn}>
+                  Profile
+                </Link>
+              </Menu.Item>
+            )}
+            <Menu.Item>
+              <Form method="post">
```
danecando, line 42: Does this work? I know the action is set to the current url by default and the navbar is on every page

remix/app/root.tsx
```diff
@@ -57,9 +66,39 @@ export function CatchBoundary() {
...
+  // verifies authed before calling to prevent redirect spiral
+  if (valid) {
+    const { accessToken } = await auth.isAuthenticated(request, {
+      failureRedirect: '/login',
+    });
+    const { viewer } = await gqlClient.request(CURRENT_USER_QUERY, undefined, {
+      authorization: `Bearer ${accessToken}`,
+    });
+    return json<LoaderData>({ viewer, valid });
+  }
```
danecando, line 93: Not needed but if we want to refactor and clean up a bit more this is a nice pattern for the session https://github.com/kentcdodds/kentcdodds.com/blob/main/app/utils/session.server.ts#L57 - Could just have a method on the session value returned from getSession to handle checking auth value etc

danecando (changes requested):

remix/app/components/UserDropdown/UserDropdown.view.tsx
WillHutt: Yes, that does work it’s using the logout action inside the root.

remix/app/root.tsx
WillHutt: I'll make a separate enhancement ticket in case we want to go this route.

WillHutt: https://github.com/thisdot/starter.dev-github-showcases/issues/251

dustinsgoodman: We should def due #251 since Kent recommends it.

remix-gql-tailwind/app/components/UserDropdown/UserDropdown.view.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <div className="py-1">
+            {username && (
+              <Menu.Item>
+                <Link to={`/${username}`} className={styles.menuBtn}>
+                  Profile
+                </Link>
+              </Menu.Item>
+            )}
+            <Menu.Item>
+              <Form method="post">
```
danecando, line 42: action="/logout"

remix-gql-tailwind/app/root.tsx
```diff
@@ -57,9 +66,39 @@ export function CatchBoundary() {
   );
 }
 
+export const action: ActionFunction = async ({ request }) => {
+  await auth.logout(request, { redirectTo: '/login' });
+};
```
danecando, line 71: Move to a resource route `/routes/logout` since the root component semantically shouldn't handle a logout action

danecando (changes requested):
So I guess we want to do the session refactor that I highlighted. Is that an upcoming PR?

danecando (approved):
LGTM

PR #261, opened 2022-05-10
[Angular-Apollo-Tailwind] Run yarn generate in amplify deploy

BrettZeidler:


BrettZeidler:
merging as it's the only way we can verify the deploy works since AWS is using the file as it appears in the `main` branch

PR #299, opened 2022-05-20
feat: setup react context

markshenouda:
Setting up react context


cra-rxjs-styled-components/src/context/UserProvider.tsx
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+  useEffect(() => {
+    request(`${GITHUB_URL_BASE}/user`)
+      .pipe(
+        tap((val) => {
+          setUser(val as IUserContext);
+        })
+      )
+      .subscribe();
```
jesus4497, line 64: Shouldn't we have a 
```javascript
return () => {
        subscription.unsubscribe();
      };
```
here? 🤔

cra-rxjs-styled-components/src/hooks/gists/use-gists.ts
```diff
@@ -1,33 +1,22 @@
...
-    });
 
   useEffect(() => {
-    getCurrentUser().subscribe((user) => {
-      setCurrentUser(user.login);
-    });
-
     const subscription: Subscription = fromFetchWithAuth<Gist[]>(
-      GISTS_URL(currentUser),
+      GISTS_URL(user?.login as string),
```
jesus4497, line 19: Instead of doing an assertion here, I would add an if statement like: if (user) {}

jesus4497 (commented):
LGTM overall!

cra-rxjs-styled-components/src/components/dropdown/Dropdown.tsx
```diff
@@ -17,38 +15,20 @@ import {
...
-      )
-      .subscribe();
-  }, []);
   return (
     <DropdownNav>
       <DropdownBtn onClick={toggleDropdown}>
         <UserDropdownAvatarContainer>
           <UserDropdownAvatarSpan>
-            <UserDropdownAvatarImage alt={avatar.alt} src={avatar.image} />
+            {user?.avatar_url && user.name && (
```
jesus4497, line 29: We should add the optional chaining for the name prop as well as this might throw an exception 
`user?.name`

markshenouda: For the first `?` tbh I added it to make typescript happy but about the user name the statement already checks on it if it's exist or not

Issue #305, opened 2022-05-24
[Next + React Query + Tailwind] Fix Previous Pagination Button Issue


iansamz:
The previous button on pagination of issues and Prs does not behave as expected. Mps likely caused by $last variable in gql.

https://www.loom.com/share/c2337c3e23c04683bd61bcf7a723c1a6

### Acceptance

- [x] Issues page fix
- [x] PR page fix
- [x] Add pagination to Org repos
- [x] Add pagination to Profile Repos

PR #309, opened 2022-05-25
[Next +  React Query + Tailwind]: #305 fix previous pagination

iansamz:
Reolves: #305 

Fix previous button problem.
Pagination has also been added to org repos and profile repos.
Design has been made more similar to github

https://www.loom.com/share/c277700e0c1647afb967cdc964fa2805


next-react-query-tailwind/src/components/UserRepos/UserRepos.data.tsx
```diff
@@ -19,9 +19,9 @@ interface UserReposProps {
 function UserRepos({ username, isOrg = false }: UserReposProps) {
   const { query } = useRouter();
```
danecando, line 20: Looks like we're not using the query param anymore

danecando (approved):
LGTM. Only comment that I have is that I was trying to build all the features as similar as possible to GitHub (which is why I used query params for filters and pagination) but I really don't think that's a huge deal not to

next-react-query-tailwind/src/components/UserRepos/UserRepos.data.tsx
iansamz: I was mimicking the RxJs pattern on angular-apollo-tailwind and the one you had previously done

dustinsgoodman: I'd prefer to keep using the query params actually and maybe upgrade the RxJS one to match this instead of this being changed.

PR #315, opened 2022-05-25
[Angular-NgRx-SCSS] Amplify deployed site fixes

lindakatcodes:
**UPDATE: Removed small fixes, to be adjusted at a later date. This PR now only makes a small adjustment for the Amplify deployed site.**

Trying to fix the deployed version of the site! Plus small fixes to the styles / data that already exist.

- The main fix is to adjust the production "apiUrl" to match what it should be. It's hard to test this with the preview deploy since it doesn't redirect to the main site - I think we'll want a new ticket / issue for changing how the redirect works, because I don't know that the preview deploys will work without it. But I think that should be a separate issue.
- Small style fix so the main login page fills the whole screen
- Small data fix so the user's list of repos is sorted by created date, which matches the other showcases

### DIFF
--- a/angular-ngrx-scss/src/app/auth/auth.component.scss
+++ b/angular-ngrx-scss/src/app/auth/auth.component.scss
@@ -1,8 +1,8 @@
 @use '../shared/styles/variables';
 
 .page {
-  width: 100vw;
-  height: 100vh;
+  width: 100%;
+  height: 100%;
   overflow: hidden;
   background: variables.$black;
   color: variables.$gray500;
--- a/angular-ngrx-scss/src/app/auth/auth.component.scss
+++ b/angular-ngrx-scss/src/app/auth/auth.component.scss
@@ -1,16 +1,16 @@
 @use '../shared/styles/variables';
 
 .page {
-  width: 100%;
-  height: 100%;
+  width: 100vw;
+  height: 100vh;
   overflow: hidden;
   background: variables.$black;
   color: variables.$gray500;
 }
 
 .form {
-  width: 90%;
-  height: 90%;
+  width: 100%;
+  height: 100%;
   display: grid;
   place-items: center;
 }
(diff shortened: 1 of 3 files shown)


angular-ngrx-scss/src/app/user/services/user.service.ts
```diff
@@ -77,7 +77,7 @@ export class UserService {
   getUserRepos(username: string): Observable<UserReposState[]> {
     const url = `${environment.githubUrl}/users/${encodeURIComponent(
       username,
-    )}/repos`;
+    )}/repos?sort='created'`;
 
     return this.http.get<UserReposApiResponse>(url).pipe(
```
Reshurum, line 82: Can we pass query params with the `this.http.get` call instead?

angular-ngrx-scss/src/app/auth/auth.component.scss
```diff
@@ -1,16 +1,16 @@
 @use '../shared/styles/variables';
 
 .page {
-  width: 100%;
-  height: 100%;
+  width: 100vw;
+  height: 100vh;
```
Reshurum, line 5: I remember running into issues with this with scrollbars on non-macs. This change used to be present but was removed. What issue is this trying to fix?

Reshurum (changes requested):

angular-ngrx-scss/src/app/auth/auth.component.scss
Reshurum: Oh I just missed a bullet point and see why this was changed. I would figure out why the parent container isn't max width of the page instead. Setting the viewport width like this is not ideal.

lindakatcodes: Ah that's a good point I didn't know about (the previous scrollbars issue). Will move this off into it's own ticket so can do more analysis of it.

angular-ngrx-scss/src/app/user/services/user.service.ts
lindakatcodes: Should be possible! Going to move this into a separate ticket so the Amplify fix can be merged in.

Issue #117, opened 2022-02-08
[Angular - NgRx - SCSS] Get gists from GitHub API


lindakatcodes:
# Get Gists from GitHub API

- [ ] Get gists from GitHub API
- [ ] Display gists in the main page
  - [ ] Name of gist
- [ ] Clicking the gist should navigate to the gist in a new tab

Issue #186, opened 2022-03-04
[Angular - NgRx - SCSS] Create gists panel


lindakatcodes:
# Create gists panel

## Background

As a user, I should see an unordered list of my gists on the main home page.

## Acceptance

- [x] Gist title
- [x] unordered list of user gists
- [x] clicking a gist opens that file in a new tab

## Notes

lindakatcodes:
I'd double check on this, but I think this is already implemented in Daian's PR.

Reshurum:
> I'd double check on this, but I think this is already implemented in Daian's PR #319

Correct

lindakatcodes:
We should still leave this ticket open - there's a few small things to fix once Daian's PR is merged that will be easier to fix in a new PR instead of continuing on the single large one.

Still to do:
- [ ] accessibility: Consider using semantic HTML elements <ul> and <li> here. 
- [ ] styling: needs mobile styling (gist section should move to bottom of page on small screens, see next-react-query deployed app for example)

For reference if needed: https://github.com/thisdot/starter.dev-github-showcases/pull/319#discussion_r912118267

PR #319, opened 2022-05-27
feat(angular-ngrx-scss): Get gists from Github API

SDaian:
### Description

- Get Gists from GitHub API

Fixes: #117 

### Screenshots

image

### DIFF
--- a/angular-ngrx-scss/src/app/home/user-gists/user-gists.component.html
+++ b/angular-ngrx-scss/src/app/home/user-gists/user-gists.component.html
@@ -1,0 +1,16 @@
+<div class="container">
+  <h3>Gists</h3>
+  <ng-container *ngIf="userGists$ | async as gists">
+    <ng-container *ngIf="gists.length > 0; else noGists">
+      <div class="list-container">
+        <ng-container *ngFor="let gist of gists">
+          <a class="link" [href]="gist.url" target="_blank">{{gist.fileName}}</a>
+        </ng-container>
+      </div>
+    </ng-container>
+  </ng-container>
+</div>
+
+<ng-template #noGists>
+  <p>User does not have any gists</p>
+</ng-template>
--- a/angular-ngrx-scss/src/app/state/user/user.actions.ts
+++ b/angular-ngrx-scss/src/app/state/user/user.actions.ts
@@ -1,4 +1,5 @@
 import { createAction, props } from '@ngrx/store';
+import { UserGistsState } from '../profile/profile.state';
 import { UserState } from './user.state';
 
 export const fetchUserData = createAction('[User API] User data requested');
@@ -10,5 +11,20 @@ export const fetchUserDataSuccess = createAction(
 
 export const fetchUserDataError = createAction(
   '[User API] User Data fetch unsuccessful',
+  props<{ error: object }>(),
+);
+
+export const getUserGists = createAction(
+  '[User API] User gists requested',
+  props<{ username: string }>(),
+);
+
+export const getUserGistsSuccess = createAction(
+  '[User API] User gists successfully received',
+  props<{ userData: UserGistsState[] }>(),
+);
+
+export const getUserGistsError = createAction(
+  '[User API] User gists unsuccessful',
   props<{ error: object }>(),
 );
--- a/angular-ngrx-scss/src/app/state/user/user.effects.ts
+++ b/angular-ngrx-scss/src/app/state/user/user.effects.ts
@@ -7,6 +7,8 @@ import {
   fetchUserData,
   fetchUserDataError,
   fetchUserDataSuccess,
+  getUserGistsError,
+  getUserGistsSuccess,
 } from './user.actions';
 
 @Injectable()
@@ -18,6 +20,18 @@ export class UserEffects {
         this.userService.getAuthenticatedUserInfo().pipe(
           map((data) => fetchUserDataSuccess({ userData: data })),
           catchError((error) => of(fetchUserDataError({ error }))),
+        ),
+      ),
+    );
+  });
+
+  loadUserGists$ = createEffect(() => {
+    return this.actions$.pipe(
+      ofType(fetchUserDataSuccess),
+      switchMap(({ userData: { username } }) =>
+        this.userService.getUserGists(username).pipe(
+          map((data) => getUserGistsSuccess({ userData: data })),
+          catchError((error) => of(getUserGistsError({ error }))),
         ),
       ),
     );
--- a/angular-ngrx-scss/src/app/user/services/user.service.ts
+++ b/angular-ngrx-scss/src/app/user/services/user.service.ts
@@ -2,6 +2,9 @@ import { HttpClient } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 import { map, Observable } from 'rxjs';
 import {
+  UserGist,
+  UserGistsApiResponse,
+  UserGistsState,
   UserOrgsApiResponse,
   UserOrgsState,
   UserReposApiResponse,
@@ -101,6 +104,21 @@ export class UserService {
           owner: {
             login: repo.owner.login,
           },
+        })),
+      ),
+    );
+  }
+
+  getUserGists(username: string): Observable<UserGistsState[]> {
+    const url = `${environment.githubUrl}/users/${encodeURIComponent(
+      username,
+    )}/gists`;
+
+    return this.http.get<UserGistsApiResponse>(url).pipe(
+      map((data) =>
+        data.map((gist: UserGist) => ({
+          url: gist.html_url,
+          fileName: Object.keys(gist.files)[0],
         })),
       ),
     );
--- a/angular-ngrx-scss/src/app/home/home.component.html
+++ b/angular-ngrx-scss/src/app/home/home.component.html
@@ -2,7 +2,12 @@
   <header class="nav">
     <app-nav-bar></app-nav-bar>
   </header>
-  <main class="main">
-    <div>Main content!</div>
-  </main>
+  <ng-container *ngIf="user$ | async as user">
+    <main class="main">
+      <aside>
+        <app-user-gists></app-user-gists>
+      </aside>
+      <div>Main content!</div>
+    </main>
+  </ng-container>
 </div>
--- a/angular-ngrx-scss/src/app/home/home.component.scss
+++ b/angular-ngrx-scss/src/app/home/home.component.scss
@@ -5,7 +5,7 @@
   grid-template-rows: auto 1fr;
   grid-template-columns: 30% 1fr;
   width: 100%;
-  height: 100%;
+  height: 100vh;
 }
 
 .nav {
@@ -14,7 +14,14 @@
 }
 
 .main {
+  display: grid;
+  grid-template-columns: 1fr 2fr;
   grid-row: 2 / 3;
   grid-column: 1 / 3;
   background: variables.$gray100;
+}
+
+aside {
+  background: variables.$white;
+  padding: 2rem;
 }
--- a/angular-ngrx-scss/src/app/home/home.component.ts
+++ b/angular-ngrx-scss/src/app/home/home.component.ts
@@ -1,5 +1,6 @@
 import { Component, OnInit } from '@angular/core';
 import { Store } from '@ngrx/store';
+import { selectUserLoginName } from '../state/user';
 import { fetchUserData } from '../state/user/user.actions';
 
 @Component({
@@ -8,6 +9,8 @@ import { fetchUserData } from '../state/user/user.actions';
   styleUrls: ['./home.component.scss'],
 })
 export class HomeComponent implements OnInit {
+  user$ = this.store.select(selectUserLoginName);
+
   constructor(private store: Store) {}
 
   ngOnInit() {
--- a/angular-ngrx-scss/src/app/home/home.module.ts
+++ b/angular-ngrx-scss/src/app/home/home.module.ts
@@ -9,6 +9,7 @@ import { ProfileNavComponent } from './profile/profile-nav/profile-nav.component
 import { ProfileAboutComponent } from './profile/profile-about/profile-about.component';
 import { ProfileReposComponent } from './profile/profile-repos/profile-repos.component';
 import { RelativeTimePipe } from '../shared/pipes/relative-time.pipe';
+import { UserGistsComponent } from './user-gists/user-gists.component';
 
 @NgModule({
   declarations: [
@@ -20,6 +21,7 @@ import { RelativeTimePipe } from '../shared/pipes/relative-time.pipe';
     ProfileAboutComponent,
     ProfileReposComponent,
     RelativeTimePipe,
+    UserGistsComponent,
   ],
   imports: [CommonModule, HomeRoutingModule],
 })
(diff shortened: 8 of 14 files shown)


Reshurum (changes requested):
Per Dustin's response in Slack we should aim for at least 80% code coverage (ideally 100%) in PRs. Could you write some tests for this new functionality?

angular-ngrx-scss/src/app/home/top-repositories/top-repositories.component.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@Component({
+  selector: 'app-top-repositories',
+  templateUrl: './top-repositories.component.html',
+  styleUrls: ['./top-repositories.component.scss'],
+})
+export class TopRepositoriesComponent {
+  username$ = this.store.select(selectUserLoginName);
+  repos$ = this.store.select(selectTopRepos);
+  profile$?: Observable<ProfileState>;
```
lindakatcodes, line 16: What is profile$ used for? I don't see it anywhere, but might be missing something.

- [ ] Is profile$ used anywhere? If not, we can remove that

angular-ngrx-scss/src/app/home/top-repositories/top-repositories.component.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  constructor(private store: Store) {}
+
+  getColor(language: string): string {
```
lindakatcodes, line 20: Also reminder note that this can be removed once the component is in place. :)

angular-ngrx-scss/src/app/home/top-repositories/top-repositories.component.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
lindakatcodes, line 1: Honestly think we can remove this test file - the component doesn't do any heavy logic so I don't think there's anything worth testing in here.

- [ ] Can remove this test file

angular-ngrx-scss/src/app/home/user-gists/user-gists.component.html
```diff
@@ -0,0 +1,16 @@
...
+      <div class="list-container">
+        <ng-container *ngFor="let gist of gists">
+          <a class="link" [href]="gist.url" target="_blank">{{gist.fileName}}</a>
+        </ng-container>
+      </div>
+    </ng-container>
+  </ng-container>
+</div>
+
+<ng-template #noGists>
```
lindakatcodes, line 14: I'm.....not sure we actually need this check? I think if there's no gists we can just leave the section blank. But we'll want to check one of the completed showcases (Angular/Apollo or Next/React-Query) from someone's account that doesn't have any gists. What do those showcases do in that case? 

- [ ] Please verify this on a completed showcase with no user gists, see if they show a message or just leave the aside blank

angular-ngrx-scss/src/app/home/user-gists/user-gists.component.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
lindakatcodes, line 1: This test file can also be deleted. The component doesn't actually handle any logic, so not much to test here.

- [ ] Can remove this test file

angular-ngrx-scss/src/app/state/user/user.actions.ts
```diff
@@ -12,3 +13,33 @@ export const fetchUserDataError = createAction(
   '[User API] User Data fetch unsuccessful',
   props<{ error: object }>(),
 );
+
+export const getUserGists = createAction(
```
lindakatcodes, line 17: Can we change these to use "fetch" instead of "get"? To keep the naming consistent throughout?

- [ ] Rename these to use "fetch" instead of "get" (all new actions)

angular-ngrx-scss/src/app/state/user/user.actions.ts
```diff
@@ -12,3 +13,33 @@ export const fetchUserDataError = createAction(
...
+  props<{ username: string }>(),
+);
+
+export const getUserTopReposSuccess = createAction(
+  '[User API] User top repos successfully received',
+  props<{ topRepos: UserReposState[] }>(),
+);
+
+export const getUserTopReposError = createAction(
+  '[User API] User top repos unsuccessful',
```
lindakatcodes, line 33: Can we change both error action types to  say "fetch unsuccessful", please? The wording without fetch feels a little off.

- [ ] Add "fetch" to both error action types, please

angular-ngrx-scss/src/app/state/user/user.effects.spec.ts
```diff
@@ -64,4 +76,101 @@ describe('UserEffects', () => {
       done();
     });
   });
+
+  it('should get the user gists from the Github API', (done) => {
+    const userStateMock: UserState = {
```
lindakatcodes, line 81: I'd recommend moving this to the top of the file and renaming it as "initialUserState" or something similar. That way you can reuse it in these checks easier. (Could also potentially import the initialUserState from the reducer or state file, whichever it lives in.

- [ ] Make this a file-wide variable you can reuse, or pull in the initialUserState variable to use instead

angular-ngrx-scss/src/app/user/services/user.service.ts
```diff
@@ -105,4 +108,49 @@ export class UserService {
       ),
     );
   }
+
+  getUserTopRepos(): Observable<UserReposState[]> {
+    const url = `${environment.githubUrl}/user/repos?sort=updated&affiliation=owner,collaborator,organization_member&per_page=20'}`;
```
lindakatcodes, line 113: Can you add a note in the REST API doc we started to explain this url, please? I think this makes good sense for top repos, so we want to capture that info there for future showcases that use the REST API.

Link to the doc: https://docs.google.com/document/d/1qrAXPiMvag1ykFCd5dl0rLV3RtcapB9Dj8HVxrnXxLc/edit

- [ ] Update REST API doc to share this url and explain the decision to use these parameters

angular-ngrx-scss/src/app/state/user/user.state.ts
```diff
@@ -10,6 +12,8 @@ export interface UserState {
   name: string;
   twitter_username: string;
   username: string;
+  repos?: UserReposState[];
```
lindakatcodes, line 15: I think this needs to be renamed to topRepos. It's possible the full list of repos will also be saved here, and that will help us know which is which and not accidentally overwrite them.

- [ ] Rename to topRepos

lindakatcodes (changes requested):
Left a few questions and requests for you. :) Let me know if there's anything that's not clear!

angular-ngrx-scss/src/app/home/top-repositories/top-repositories.component.html
```diff
@@ -0,0 +1,42 @@
+<div class="top-repositories-container">
+  <h2>Top Repositories</h2>
+  <div class="repo-container">
+    <ng-container *ngIf="repos$ | async as repos">
+      <ng-container *ngFor="let repo of repos; last as isLast">
+        <div class="repo-section">
```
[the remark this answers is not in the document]
SDaian: agree, @lindakatcodes we should use the `app-repo-card` when is ready, can take of that if that PR either is merged before or after this PR.

angular-ngrx-scss/src/app/home/top-repositories/top-repositories.component.scss
```diff
@@ -0,0 +1,98 @@
...
+    line-height: 28px;
+    font-weight: 500;
+    margin-bottom: 16px;
+  }
+  .repo-container {
+    background-color: variables.$white;
+    border-radius: 0.5rem;
+    border-width: 1px;
+  }
+  .repo-section {
```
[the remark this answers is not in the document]
SDaian: same as above, will be cleaned up once one of the PR was merged.

angular-ngrx-scss/src/app/state/user/user.effects.ts
```diff
@@ -23,5 +27,29 @@ export class UserEffects {
     );
   });
 
+  loadUserGists$ = createEffect(() => {
+    return this.actions$.pipe(
+      ofType(fetchUserDataSuccess),
```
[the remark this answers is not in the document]
SDaian: yes indeed, is the only trigger to get this data from the API.

angular-ngrx-scss/src/app/home/top-repositories/top-repositories.component.html
```diff
@@ -0,0 +1,43 @@
...
+          <h3>
+            <a class="name" href="/{{ repo.owner.login }}/{{ repo.name }}">{{
```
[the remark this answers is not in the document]
lindakatcodes: This is already handled in the shared component that will be replacing this, so not needed here. :)

angular-ngrx-scss/src/app/home/top-repositories/top-repositories.component.html
```diff
@@ -0,0 +1,43 @@
...
+            </div>
+            <div *ngIf="repo.license" class="license">
+              <span class="icon" appOcticon="law"></span>
+              <span class="license-name">{{ repo.license.name }}</span>
+            </div>
+            <div class="updated">{{ repo.updated_at | relativeTime }}</div>
+          </div>
+        </div>
+        <div class="view-all-link" *ngIf="isLast">
+          <a routerLink="/{{ username$ | async }}">View all repositories</a>
```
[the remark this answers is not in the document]
lindakatcodes: As this is at the bottom of the page and a selector, this is what we'd want. It's the only way we'd have the username. Should be just fine by the time someone gets down to the bottom of the page.

angular-ngrx-scss/src/app/state/user/user.state.ts
```diff
@@ -10,6 +12,8 @@ export interface UserState {
   name: string;
   twitter_username: string;
   username: string;
+  topRepos?: UserReposState[];
+  gists?: UserGistsState[];
```
Reshurum, line 16: suggestion: I think these should be attached to home page specific state rather than being a part of the global user state; however I know we discussed possibly refactoring some of the state management in this showcase, so we can elect to fix that later if that makes more sense.

angular-ngrx-scss/src/app/user/services/user.service.spec.ts
```diff
@@ -55,4 +62,182 @@ describe('UserService', () => {
...
+          starred_url:
+            'https://api.github.com/users/thisdot/starred{/owner}{/repo}',
+          subscriptions_url:
+            'https://api.github.com/users/thisdot/subscriptions',
+          type: 'Organization',
+          url: 'https://api.github.com/users/thisdot',
+        },
+      },
+    ];
+    httpClientSpy.get.and.returnValue(of(expectedHttpResponse).pipe(delay(0)));
```
Reshurum, line 211: question: Why is this `.pipe(delay(0))` needed?

angular-ngrx-scss/src/app/user/services/user.service.spec.ts
```diff
@@ -55,4 +62,182 @@ describe('UserService', () => {
...
+        html_url: 'github.com/gists',
+        files: { 'textfile1.txt': { filename: 'textfile1.txt' } },
+      },
+    ];
+
+    httpClientSpy.get.and.returnValue(of(expectedHttpResponse).pipe(delay(0)));
+
+    userService.getUserGists(username).subscribe((res) => {
+      expect(res).toEqual(expectedResponse);
+      done();
```
Reshurum, line 240: nitpick: We should call `done()` inside of the `complete` callback of the subscription instead of the success one. If an error happens in the observable, this test may hang and never complete.

angular-ngrx-scss/src/app/user/services/user.service.ts
Reshurum, line 113: suggestion: GET params should be included using the params object that the HTTP service accepts. I would consider implementing params in a manner similar to [this snippet](https://github.com/thisdot/starter.dev-github-showcases/blob/8648d0c18a87e61c460cbcf552ba5d9fd5ffddaf/angular-ngrx-scss/src/app/organization/services/organization.service.ts#L16-L40).

You could forgo accepting the params as a parameter in the function since this call is ad-hoc, but I would still recommend keeping these values in a map so they are easier to read and edit. You could pass a literal map to the `params` key in this case and that should work. You do not need to define an interface in this context like in the aforementioned example if I recall correctly.

Reshurum (changes requested):

angular-ngrx-scss/src/app/user/services/user.service.spec.ts
Reshurum: I haven't tested this though. Maybe subscriptions work differently in tests.

angular-ngrx-scss/src/app/home/user-gists/user-gists.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="container">
+  <h3>Gists</h3>
+  <ng-container *ngIf="userGists$ | async as gists">
+    <ng-container *ngIf="gists.length > 0">
+      <div class="list-container">
```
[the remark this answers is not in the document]
lindakatcodes: I agree with you, but wondering if it might be worth making that adjustment in a separate PR, since this one is already so over scope and I know we have the ticket for adding this view still open? #186 

The gists panel also doesn't have the mobile stylings here (to match the NextJS version of the site), so both those updates could be handled in that ticket instead.

angular-ngrx-scss/src/app/state/user/user.state.ts
lindakatcodes: I would like to make these adjustments as part of a larger overhaul to how the store is set up, as we discussed. I think it's worth leaving them for now and making a new ticket to correct the state structures all over. Will keep those changes easier to review, as well, I think.

angular-ngrx-scss/src/app/home/user-gists/user-gists.component.html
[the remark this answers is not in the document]
Reshurum: So long as we leave a note of this somewhere that can be found, I am fine with this being done later (personally).

angular-ngrx-scss/src/app/state/user/user.state.ts
Reshurum: Ok! Makes sense to me.

angular-ngrx-scss/src/app/home/top-repositories/top-repositories.component.html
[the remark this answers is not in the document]
lindakatcodes: Couldn't quite make this into a getter, but did create a small subscription (with a take to complete it once we get the value) so the url isn't formed in the template.

angular-ngrx-scss/src/app/user/services/user.service.spec.ts
SDaian: Removed, since now the observables are handled from the fakeAsync approach.

angular-ngrx-scss/src/app/user/services/user.service.spec.ts
SDaian: Completely agree with you @Reshurum, doing a little bit of research found this quite interesting article who mention we should avoid the `done()` method on the spec. because can cause some false positives and a slow performance on tests:

[https://www.piotrl.net/angular-testing-avoid-done/](https://www.piotrl.net/angular-testing-avoid-done/)

So basically I rewrite using `fakeAsync` to handle the Observables.

lindakatcodes (approved):

PR #332, opened 2022-06-02
Get Organization repositories from Github API

vyktoremario:
Screenshot 2022-06-02 at 10 17 17
Screenshot 2022-06-02 at 10 24 46


cra-rxjs-styled-components/src/components/repo-meta/RepoMeta.tsx
```diff
@@ -0,0 +1,63 @@
...
+            </SocialCount>
+          )}
+          {forkCount > 0 && (
+            <SocialCount>
+              <BranchLogo />
+              <TextSpan>{forkCount}</TextSpan>
+            </SocialCount>
+          )}
+        </SocialWrapper>
+      )}
```
dustinsgoodman, line 52: I find this to be a bit hard to read because of all the embedded logic. Is there a way we could extract this to be cleaner?

cra-rxjs-styled-components/src/components/user-repos/UserRepos.tsx
```diff
@@ -0,0 +1,102 @@
...
+} from './UserRepos.styles';
+import RepoMeta from '../repo-meta';
+import PrivacyBadge from '../misc/Privacy-badge';
+
+function UserRepos({ isOrg = false }) {
+  const [repos, setRepos] = useState<Repo[]>();
+  const [loading, setLoading] = useState(true);
+  const params = useParams();
+
+  const GITHUB_URL = isOrg ? ORG_REPO_LIST(params.username!) : ''; // The empty string can be replaced with URL to get all User Repos.
```
dustinsgoodman, line 28: I don't like the non-null assertion operator in TypeScript. It's usually hiding a code smell. I also don't understand the empty string thing and why we're doing that here.

cra-rxjs-styled-components/src/components/user-repos/UserRepos.tsx
```diff
@@ -0,0 +1,102 @@
...
+      .pipe(
+        tap((val) => {
+          if (val) {
+            setRepos(val);
+            setLoading(false);
+          }
+        })
+      )
+      .subscribe();
+  }, [GITHUB_URL, params.username]);
```
dustinsgoodman, line 48: Should we abstract this into its own hook?

dustinsgoodman (changes requested):
There's both a package-lock.json and yarn.lock. We need to pick a single package manager for this project. I think the kit uses npm so let's stick to that please.

cra-rxjs-styled-components/src/components/repo-meta/RepoMeta.tsx
vyktoremario: Yea, that's possible.

vyktoremario:
> There's both a package-lock.json and yarn.lock. We need to pick a single package manager for this project. I think the kit uses npm so let's stick to that please.

I actually thought we used yarn. I will delete the `yarn.lock` file.

cra-rxjs-styled-components/src/components/user-repos/UserRepos.tsx
vyktoremario: Alright. This make sense

cra-rxjs-styled-components/src/hooks/user-repositories/use-user-repositories.ts
```diff
@@ -0,0 +1,43 @@
...
+    request(GITHUB_URL)
+      .pipe(
+        tap((data) => {
+          if (data) {
+            setRepos(data);
+            setLoading(false);
+          }
+        })
+      )
+      .subscribe();
```
dustinsgoodman, line 36: Don't forget to unsubscribe in the umount.

cra-rxjs-styled-components/src/components/repo-meta/RepoMeta.tsx
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+      )}
+      {(stargazerCount > 0 || forkCount > 0) && (
+        <SocialWrapper>
+          {stargazerCount > 0 && (
+            <MetaIcon text={stargazerCount} children={<StarLogo />} />
+          )}
+          {forkCount > 0 && (
+            <MetaIcon text={forkCount} children={<BranchLogo />} />
+          )}
```
dustinsgoodman, line 40: So I still have an issue with this specifically... I don't like having logic wrapping logic for rendering. I'd almost prefer before the return to build the social icons and then inject them here instead or defer some of the render logic down a layer. So we could do something like:

```
const socialIcons = [];
if (stargazerCount > 0) {
  socialIcons.push(<MetaIcon text={stargazerCount} children={<StarLogo />} />);
}
if (forkCount > 0) {
  socialIcons.push(<MetaIcon text={forkCount} children={<StarLogo />} />);
}

return (
   // ...
  {socialIcons.length > 0 && (<SocialWrapper>{socialIcons}</SocialWrapper>)}
);
```

or something that moves the logic out of the view like this.

vyktoremario: I think this is neat

dustinsgoodman (approved):

cra-rxjs-styled-components/src/components/repo-meta/RepoMeta.tsx
dustinsgoodman: Glad you liked it! It's mostly a readability thing. I have the same issue with nested ternaries. If I see layered logic, I see a smell for someone later on. What's cool about this too is we could theoretically add another section if it exists.

Issue #337, opened 2022-06-06
[CRA-RxJS-SC] Refractor Repo sub-header & tab navigation


iansamz:
Refractor and fix

- [x] Refractor sub-header and route outlet for repos
- [x] issues on tab navigation showing wrong numbers
- [x] Tab navigation when active

vyktoremario:
@iansamz I think a screenshot or a video of the issue with reference to the actual behaviour will give this more context

iansamz:
Noted. Updating

PR #338, opened 2022-06-06
[CRA-RxJS-SC] Refractor Repo sub-header & tab navigation

iansamz:
Resolves: #337 

1. I refactored the routing in `index.tsx` for the individual repo page. Allowing for routing for PRs, issues and code, under one AuthGuard.
2. I refactored the sub-header to correctly get the right number of issues using `https://api.github.com/search` instead of `/issues` as used before. This also allows for future  pagination and filtering for PRs and issues.
3. I added functionality for active tabs based on routes in the subheader
4. Fix: Changed all fill-rule to fillRule for SVGs

https://www.loom.com/share/7390f113d8e6431b8ad307b45878d31c


cra-rxjs-styled-components/src/components/sub-header/SubHeader.tsx
```diff
@@ -29,32 +28,95 @@ import {
...
+  const request = (url: string) =>
+    fromFetchWithAuth(url, {
+      selector: (response: Response) => {
+        return response.json();
+      },
+    });
+
+  useEffect(() => {
+    forkJoin([
+      request(SINGLE_USER_REPO(params.username!, params.repo!)),
```
vyktoremario, line 69: It looks like most of these implementations will be repeated again in the individual `repository-code`, `repository-issues` and `repository-pull-request`.

vyktoremario: Also, I am sure this data is fetched here and is just used to get issues and PR counts?

vyktoremario (changes requested):
Not so sure I understand or like the code structure here. 

I feel we will need a context and that will provide some data to its children. 
So instead of fetching data in the SubHeader Component just to get count, you can receive that data from a context. 
I believe in that way, we are just making a single call or calls to the endpoints and the data are distributed accordingly to the subHeader component to get the issues and PR count and also to the respective routes to use and display the data. 

I don't know if this makes sense. Ping me on slack if you need a hand with this.

cra-rxjs-styled-components/src/components/sub-header/SubHeader.tsx
vyktoremario: I don't like the non-null assertion operator in TypeScript. It's usually hiding a code smell.

iansamz: I used this branch as a parent for pr #339

iansamz: > Also, I am sure this data is fetched here and is just used to get issues and PR counts?

The data used is for subheader data like issues count and pr count.  The data from the issues had wrong counts as github counts PRs as issues.

PR #362, opened 2022-06-10
[Cypress] Add organization page cypress tests

### LINKED ISSUE #308: [Cypress] Write test coverage for organization page
Tests may potentially share with #265, but we will want to:

- Check that the organization information is correct at the top of page
- Repositories are correct to that organization and not to the user's repositories

BrettZeidler:
Closes #308


jnalei (approved):
Nice cleanup for the interceptors!

PR #395, opened 2022-06-20
[Angular-NgRx-SCSS] Implement Reusable RepoCardComponent

### LINKED ISSUE #187: [Angular - NgRx - SCSS] Create repository card
# Create repository card

## Background

This will be a reusable component that displays details about a repository. On the home page and profile page, a user will see a list of their repositories. This component is intended to be the building block for that - a single card for a repository, showing pertinent data about that repo.

## Acceptance

- [ ] Card view for use on repository lists
- [ ] should always have:
    - [ ] repo name
    - [ ] visibility tag (public / private)
    - [ ] description (if available)
    - [ ] main language with color dot
    - [ ] star icon with number of stars on repo
    - [ ] last updated text
    - [ ] optional - star button on right to add a star to the repo (does not need to function yet)

## Notes

Reshurum:
Resolves #187


angular-ngrx-scss/src/app/shared/components/repo-card/repo-card.component.html
```diff
@@ -0,0 +1,27 @@
+<ng-container *ngIf="repo">
+  <h3 class="name-container">
+    <a class="name" href="/{{ repo.owner.login }}/{{ repo.name }}">{{
```
ihardz, line 3: We can place the logic (for example url composing logic) into component code and just bind it. Otherwise the template has alot responsibilities. I call this as Single Responsibility Principle.  From my point of view the template should only display the data.

angular-ngrx-scss/src/app/shared/components/repo-card/repo-card.component.html
```diff
@@ -0,0 +1,27 @@
...
+      repo.name
+    }}</a>
+    <span class="visibility">{{ repo.private ? 'Private' : 'Public' }}</span>
```
ihardz, line 6: I would propose to move logic into component code. As a public getter property

angular-ngrx-scss/src/app/shared/components/repo-card/repo-card.component.scss
```diff
@@ -0,0 +1,64 @@
...
+      height: 12px;
+      border-radius: 50%;
+      margin-right: 8px;
+    }
+  }
+
+  .stars {
+    display: flex;
+    align-items: center;
+    line-height: 16px;
```
ihardz, line 52: use **em**-s but not an absolute value. Bcz if you will change the font size the look will be inconsistent

angular-ngrx-scss/src/app/shared/pipes/relative-time.pipe.spec.ts
```diff
@@ -1,8 +1,57 @@
...
+  });
+
+  it('should report years if >=12 months (plural)', () => {
+    expect(pipe.transform(mockDatetime(2 * YEAR))).toBe('2 years ago');
+  });
 });
+
+function mockDatetime(seconds: number): string {
+  const now = new Date().getTime();
+  return new Date(+now - seconds * 1000).toISOString();
```
ihardz, line 56: now is already number. remove excess casting with **+**

lindakatcodes (changes requested):
Looks pretty good! Just one small change request in the repo template / component, and a question in the spec. :)

lindakatcodes (approved):
Think this should be good to go!

Reshurum:
> Think this should be good to go!

Thanks! It seems like a test fails in CI mode only. I am going to try to fix that before merging this if that's okay.

PR #412, opened 2022-06-22
chore: upgrade remix to latest

WillHutt:


danecando (changes requested):
Can you run the codemod `npx @remix-run/dev migrate` it will automatically fix imports for a change in 1.6.0

WillHutt:
Hmm, I tested everything, it seems fine but I will do so.

danecando:
I think they are going to remove imports from 'remix' in a future update. The codemod changes the imports to explicitly import from each actual package ie `@remix-run/react`

danecando (approved):
sweet 🚀

Issue #329, opened 2022-05-31
[CRA-RxJS-SC] create file view


dustinsgoodman:
## Background

As a user, if I’m viewing a single repository and click on an individual file, I am navigated to a single page view of that file. This should look similar to how a file would look on GitHub itself.

## Acceptance

- [x] clicking on a file in a repo displays that file

## Notes

PR #425, opened 2022-06-27
[CRA + RXJS + SC] Create File View

iansamz:
Resolves: #329


cra-rxjs-styled-components/src/routes/repo/repository-code/repository-blob/repository-blob.tsx
```diff
@@ -0,0 +1,17 @@
+import FileViewer from '../../../../components/file-viewer';
+import {
+  RepoContainer,
+  RepoCenterWrapper,
+} from '../../../../components/layouts/RepoLayoutPage';
+
+export default function RepoBranchBlobPath() {
+  return (
+    <>
```
danecando, line 9: Just a nitpick but you don't need a fragment here

danecando (dismissed):
Not sure if the file/repo crumbs are part of a different ticket but just in case I'm raising it since I noticed it
Screen Shot 2022-07-15 at 12 15 35 PM

One style thing is the file viewer container doesn't have rounded corners on yours but other then that it looks good to me. Add a border radius and I think you nailed it sir

markshenouda (approved):

PR #432, opened 2022-06-30
[Angular-NgRx-SCSS] Adjust login logic

### LINKED ISSUE #331: [Angular - NgRx - SCSS] Adjust log in logic to persist authentication status
Because the urls for the deploy preview and the signIn redirect (which is hard coded in the env variables) are different, the NgRx store doesn’t persist across them, and so the calls to set a user as authenticated and to store the session cookie don’t work. It’s fine locally and on the fully deployed instance since those urls will match, but will prevent the previews from letting anyone log in. 

Will need some discussion on how best to adjust for this for deploy previews:
- Get the current url dynamically, and use that instead of hard coding the url paths?
- Move the logic that saves the session cookie to a route guard or something similar, and adjust the store logic to check for the cookie first?
- Move all auth logic out of the NgRx store and make it local / cookie based?

Not sure the best path forward. But the deploy previews would be helpful enough for reviewing that finding a solution for this is worth figuring out.

Edit: Looking into all these possibilities, but adding in some acceptance criteria for the end result.

**Acceptance criteria:**
- Sign in should be possible locally, through deploy previews, and through the deployed app
- If a user is signed in, they should stay signed in after a page refresh

lindakatcodes:
Resolves #331 

Adjusted the login logic to both:
- Allow PR previews to work, by checking the url host and adjusting the redirect url if it matches the pr preview url
- Adjusting the auth guard to check for the session cookie instead of relying directly on the store, so page refreshes while a user is still logged in will work

Also adjusted the related tests to work with these changes.

### DIFF
--- a/angular-ngrx-scss/src/app/auth/redirect/redirect.component.ts
+++ b/angular-ngrx-scss/src/app/auth/redirect/redirect.component.ts
@@ -2,7 +2,7 @@ import { Component, OnInit } from '@angular/core';
 import { Router } from '@angular/router';
 import { Store } from '@ngrx/store';
 import { take } from 'rxjs';
-import { loadUserToken, selectIsAuthenticated } from 'src/app/state/auth';
+import { saveUserToken, selectIsAuthenticated } from 'src/app/state/auth';
 @Component({
   selector: 'app-redirect',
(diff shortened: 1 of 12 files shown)


SDaian (approved):
+1

PR #434, opened 2022-07-01
[Angular + Apollo + Tailwind] Fix loading state

### LINKED ISSUE #58: [Angular + Apollo + Tailwind] Fix loading state
Some parts of the app have inconsistent content loading patterns. For instance, Issues and Pull Requests handle loading state through StoreComponent logic so there's more control, but apollo requests have loading state mechanisms built in that don't work quite right.

Use inspiration from this [example](https://github.com/wescopeland/this-dot-exercise/blob/master/src/app/user-search/user-search.component.ts), we want to control the loading state so the appropriate `content-loader` components and error states render.

- [x] either find a way to properly use the `sd-data-container` component or deprecate it
- [x] add loading state control on all other queries _except_ issues and pull requests
- [x] #444
- [x] remove the `filter` project lib back into the main source code. this piece is complex but doesn't need to be its own project
- [x] cleanup `sd-container` usage(s) across the site (if not already handled)

ihardz:
Resolves #58

### DIFF
--- a/angular-apollo-tailwind/projects/shared/src/lib/components/data-container/data-container.module.ts
+++ b/angular-apollo-tailwind/projects/shared/src/lib/components/data-container/data-container.module.ts
@@ -2,6 +2,9 @@ import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
 import { DataContainerComponent } from './data-container.component';
 
+/**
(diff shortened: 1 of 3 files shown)


morgnism (approved):
LGTM

PR #438, opened 2022-07-01
[Vue3 Apollo Quasar] add search functionality

WillHutt:
# General

### About

- Implements search functionality

### Screenshot

image


WillHutt:
@AllanJeremy @hdJerry I was able to implement the functionality, but I have some questions/need some guidance on the input component. 

I noticed that the `AppInput component` emits an event that we get access to in the `SearchInput component`. I could not figure out how to listen/emit the event further so to test my filter implementation I just put a basic input type into the `SearchFilter component`. Obviously, this needs to be reverted to use the `SearchInput component` again.

How do I go about emitting/listening to the `AppInput component` event? 
It needs to go from: 
`AppInput component` -> `SearchInput component` -> `SearchFilter component` -> `ProfilePageLayout component`

hdJerry:
> @AllanJeremy @hdJerry I was able to implement the functionality, but I have some questions/need some guidance on the input component.
> 
> I noticed that the `AppInput component` emits an event that we get access to in the `SearchInput component`. I could not figure out how to listen/emit the event further so to test my filter implementation I just put a basic input type into the `SearchFilter component`. Obviously, this needs to be reverted to use the `SearchInput component` again.
> 
> How do I go about emitting/listening to the `AppInput component` event? It needs to go from: `AppInput component` -> `SearchInput component` -> `SearchFilter component` -> `ProfilePageLayout component`

HI @WillHutt 
You don't have to worry about what the emit is doing. But what it does though is to ensure that where the component `AppInput` is used you can access its value with the v-model attribute which has ref `search`.So, what you can do now with that ref is to use watchEffect that with help track any changes of the ref

Like this in the searchinput component 
```js
import { defineComponent, ref, watchEffect } from 'vue';
import { AppInput } from '../';

export default defineComponent({
  name: 'SearchInput',
  components: { AppInput },
  setup() {
    const search = ref('');

    watchEffect(() => console.log(search.value));

    return { search };
  },
});

```

vue3-apollo-quasar/src/components/ProfilePageLayout/ProfilePageLayout.vue
```diff
@@ -86,8 +86,32 @@ function changeTab(val) {
...
+const SEARCH_QUERY = gql`
+  query Search {
+    search @client
+  }
+`;
+
+const { result: searchData, loading: loadingSearch } = useQuery(SEARCH_QUERY);
 const user = useUserStore();
 const { repos, loading } = getUserRepos(props.username, false);
+const filteredRepos = computed(() => {
```
hdJerry, line 98: You can simplify this with fewer code and get same result

```js
const filteredRepos = computed(() => {
  return repos.value.filter((repo) => {
    return repo.name.match(new RegExp(searchData.value.search, 'i'));
  });
});
```

filter with return an array and with the match stuff using the flag `i` you don't have to worry about case sensitivity.

hdJerry (changes requested):

vue3-apollo-quasar/src/components/ProfilePageLayout/ProfilePageLayout.vue
WillHutt: Ah! I took that logic from a different starter project(next-react-query), and it's used in many others. Do we want to change logic away from the existing stack? As is, I don't see a reason for this to hold up the merge. 😃

hdJerry: I understand. Just pointing out the simplicity and its cleaner. :smile:

hdJerry (approved):

PR #441, opened 2022-07-02
[Angular - NgRx - SCSS] Repo controls shared component

### LINKED ISSUE #189: [Angular - NgRx - SCSS] Create component for search input and filter / sort buttons
#  Create component for search input and filter / sort buttons

## Background

This is intended to be a reusable component, wrapping around a few different items that will have their functionality added in another ticket. The focus of this ticket is to create the items themselves and a wrapper around them, so the section can be used in multiple pages. This component will contain an input field for search queries, and a series of dropdown buttons for filtering and sorting. 

## Acceptance

- [x] component will contain the following items - for design of pieces only, do not need functionality yet
    - [x] search input with “find a repository..” placeholder
    - [x] type dropdown button (does not need to contain options yet)
    - [x] language dropdown button (does not need to contain options yet)
    - [x] sort dropdown button (does not need to contain options yet)
    - [x] divider line
    - [x] space for showing text relating to selected options
    - [x] clear filter button with x icon (does not need functionality yet)

## Notes

lindakatcodes:
Resolves #189 

Added this shared component to the profile page (which is one location where it would be used) so it's visible for review. Screenshot is also attached below.

Does not include any of the functionality (as the ticket suggested it doesn't need), but form control is set up for the input. The results text is just hard coded text as well, just as an FYI.

Tried to match it as best I could to the Next / Reach Query deployed site. Some things are a touch different, but should be fairly close.

Screen Shot 2022-07-01 at 7 18 53 PM


SDaian:
LGTM! Nice job @lindakatcodes.

Resolve the conflicts and should be ready to go :)

SDaian (approved):
+1

PR #443, opened 2022-07-05
Icon resolution vue3 apollo quasar

hdJerry:
- Quasar custom icon
- Updated Readme on how to use the custom icon


vue3-apollo-quasar/src/assets/book.svg
```diff
@@ -0,0 +1,4 @@
+<svg height="1rem" viewBox="0 0 16 16" version="1.1" width="1rem" fill="none"
```
AllanJeremy, line 1: Missing `id` for this one

AllanJeremy (changes requested):
LGTM, one minor detail, the book icon seems to be missing an `id`

vue3-apollo-quasar/src/assets/book.svg
AllanJeremy: Ah, okay! In that case, please do delete it :)

AllanJeremy (changes requested):

PR #452, opened 2022-07-08
[Cypress] Add empty list tests for dashboard

### LINKED ISSUE #418: [Cypress] Add empty list tests for Dashboard page

jnalei:
PR closes #418 
Add empty list tests for dashboard page


starter-dev-e2e/cypress/support/e2e.ts
```diff
@@ -1,13 +1,10 @@
...
-import { View } from "../utils/view";
 
 beforeEach(() => {
   cy.intercept(/.*\/auth.token/, {
     fixture: "auth/token.json",
   }).as("token");
   cy.mockNextAuthCookie();
   cy.mockRemixAuthCookie();
-  cy.interceptGraphQLCalls(View.Dashboard);
-  cy.interceptRestCalls(View.Dashboard);
```
BrettZeidler, line 11: Confirmed this isn't needed -- in `dashboard.cy.ts` in the `beforeEach` let's add the following on line 8: `cy.visit("/");`

This will cause calls that were failing on the redirect to this page in this `e2e` file to be recalled and allow the flaky intercepts due to a race condition happen reliably!

starter-dev-e2e/cypress/e2e/dashboard.cy.ts
```diff
@@ -48,3 +49,36 @@ describe("When there is proper dashboard page responses", () => {
...
 });
+
+describe("When there is proper empty dashboard page responses", () => {
+  beforeEach(() => {
+    cy.interceptGraphQLCalls(View.Dashboard, InterceptResponse.Empty);
+    cy.interceptRestCalls(View.Dashboard, InterceptResponse.Empty);
+  });
+
+  // Error displayed instead of empty list.
+  // Angular-Apollo has empty message but displays error; Next-React empty & error msg displays same
```
BrettZeidler, line 60: Yeah this seems to be a bug in both -- let's write the tests for the desired behavior and let it be filed as a bug against the showcases. I assume all we need to do is uncomment below?

BrettZeidler (changes requested):
Looks great, left a couple small comments and it will be good to go!

starter-dev-e2e/cypress/e2e/dashboard.cy.ts
jnalei: Yes, I used the empty gists text from the angular-apollo project.

BrettZeidler (approved):

Issue #197, opened 2022-03-04
[Angular - NgRx - SCSS] Create single repo view about section


lindakatcodes:
# Create single repo view about section

## Background

A sub-component of the single repo page, this component focuses on the about section. It should show the user details about the repository, and any links made available to the user from that section.

## Acceptance

- [x] About title
- [x] repo description
- [x] link icon and url for deployed version (if available)
- [x] open book icon and readme title
- [x] divider line

PR #462, opened 2022-07-13
[Angular - NgRx - SCSS] Create single repo view about section

asincole:
A sub-component of the single repo page, this component focuses on the about section. It should show the user details about the repository, and any links made available to the user from that section.

Closes: #197


angular-ngrx-scss/src/app/file-viewer/file-explorer-root/file-explorer-root.component.ts
```diff
@@ -0,0 +1,21 @@
...
+  templateUrl: './file-explorer-root.component.html',
+  styleUrls: ['./file-explorer-root.component.scss'],
+})
+export class FileExplorerRootComponent implements OnInit {
+  repo$ = this.store.select(selectRepositoryState);
+
+  constructor(private route: ActivatedRoute, private store: Store) {}
+
+  ngOnInit() {
+    const owner = this.route.snapshot.paramMap.get('owner') as string;
```
lindakatcodes, line 17: suggestion: We should be able to have this data already in the store. Can we not pull it from there, rather than grabbing values from the route?

- [ ] Use store to get owner and repo name instead of router

angular-ngrx-scss/src/app/state/repository/repository.selectors.ts
```diff
@@ -1,6 +1,6 @@
 import { createFeatureSelector } from '@ngrx/store';
 import { RepoState } from './repository.state';
 
-export const repositoryFeatureKey = 'repository';
+export const repositoryFeatureKey = 'repo';
```
lindakatcodes, line 4: suggestion: What's the reason for changing the feature key? I don't think this needs to be changed.

- [ ] Why is this adjusted?

lindakatcodes (changes requested):
Overall I think the look of everything is great in the app! 

But I have some concerns about the structure of the site, and how much of that is changing in this PR that's not entirely related to that (though I also can see where you were coming from for a lot of it).

The biggest thing is: 
BLOCKER: Why is the home routing adjusted? I'd think what we actually want is to have a new "page" for the single repo view, instead of completely adjusting the home page (which is meant to be the main dashboard, aka the gists and top repos)

In the changes you made, some clarity is needed around:
- What section is the "repository-details" component supposed to be? Can you share a screenshot?
- Top repositories doesn't seem to fit in the new setup, as it's both top repos and gists, which is two separate things. 

I left a few other comments in more specific places, nothing huge. This, I think, will be the biggest thing to decide as a team, and then adjust as needed so we keep things clean and consistent. Please let me know if anything needs clarification.

angular-apollo-tailwind/src/app/file-viewer/file-explorer-root.component.spec.ts
```diff
@@ -1,14 +1,15 @@
...
+import { FileExplorerAboutComponent } from './components/file-explorer-about/file-explorer-about.component';
 
 describe('FileExplorerRootComponent', () => {
   let component: FileExplorerRootComponent;
   let fixture: ComponentFixture<FileExplorerRootComponent>;
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
-      declarations: [FileExplorerRootComponent],
+      declarations: [FileExplorerRootComponent, FileExplorerAboutComponent],
```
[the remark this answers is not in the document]
asincole: I ran tests and I was getting warning logs (even though the tests passed). I'm just sort of a fan of clean logs as much as possible 😭

angular-ngrx-scss/src/app/state/repository/repository.selectors.ts
asincole: This was changed because the state slice is called "repo", so the initial version was not working till I updated the key.

angular-ngrx-scss/src/app/file-viewer/file-explorer-root/file-explorer-root.component.ts
asincole: I think I may be a bit confused here. Can we look through this via a pair session?

angular-ngrx-scss/src/app/file-viewer/components/file-explorer-about/file-explorer-about.component.ts
```diff
@@ -0,0 +1,24 @@
+import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
+import { RepoState } from '../../../state/repository';
+
+@Component({
+  selector: 'app-file-explorer-about',
+  templateUrl: './file-explorer-about.component.html',
+  styleUrls: ['./file-explorer-about.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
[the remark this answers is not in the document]
asincole: This is pretty much a dumb component, so the data it uses comes from a parent component. This makes it such that angular does not run change detection on this component except in cases where the data supplied changes (helps with perf when you have lots of components)

angular-ngrx-scss/src/app/file-viewer/components/file-explorer-about/file-explorer-about.component.ts
```diff
@@ -0,0 +1,24 @@
...
+})
+export class FileExplorerAboutComponent {
+  @Input() repo!: RepoState | null;
+
+  get description(): string | null | undefined {
```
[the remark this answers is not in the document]
asincole: Following the point above (the dumb component one), the data is gotten already by an ancestor (which does so using a selector), so all this component needs is for that to pass the data down to it.
The getters may be a slight overkill since I could simply rely on the 'repo' prop and use null checks in the template, however, I wanted a bit of consistency.

angular-ngrx-scss/src/app/file-viewer/components/file-explorer-about/file-explorer-about.component.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    fixture = TestBed.createComponent(FileExplorerAboutComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should render "about" title', () => {
```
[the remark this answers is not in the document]
asincole: haha. thanks lol

angular-ngrx-scss/src/app/file-viewer/components/file-explorer-about/file-explorer-about.component.html
```diff
@@ -0,0 +1,29 @@
...
+
+    <ng-container *ngIf="topics">
+      <div class="mt-1">
+        <span *ngFor="let topic of topics" class="topic">{{ topic }}</span>
+      </div>
+    </ng-container>
+  </div>
+
+  <div class="mt-1">
+    <!--    TODO: this should scroll down to the readme section of the repository when it is clicked-->
```
morgnism, line 23: TODO's tend to get lost in the code. Be sure to add these to a new issue if not created already.

angular-ngrx-scss/src/app/file-viewer/file-explorer-root/file-explorer-root.component.html
```diff
@@ -0,0 +1,10 @@
+<div class="container grid grid-cols-12 subpage">
+  <section class="col-span-9">
+    <!--    TODO: replace the text below with actual repo view-->
```
morgnism, line 3: Same comment here for TODO

angular-ngrx-scss/src/app/repository/components/repository-details/repository-details.component.html
```diff
@@ -0,0 +1,4 @@
+<ng-container>
+  <!--  TODO: the repo header component should go here when created -->
```
morgnism, line 2: Same comment here for TODO.

PR #464, opened 2022-07-14
Vue3 apollo quasar/feat/type filter functionality

hdJerry:


vue3-apollo-quasar/src/components/ProfilePageLayout/ProfilePageLayout.vue
```diff
@@ -99,15 +104,36 @@ function changeTab(val) {
...
+  }
+`;
+
+const { result: filterTypeData, loading: loadingFilterType } =
+  useQuery(FILTER_TYPE_QUERY);
+
+const { repos, loading } = getUserRepos(props.username, false);
+
+const filteredTypeRepoData = computed(() => {
+  let resp: any = [];
```
AllanJeremy, line 119: `resp` seems vague as a variable name.
perhaps name it something more idiomatic: eg. `response`

Also, can we avoid using `any` as a type?

vue3-apollo-quasar/src/components/SearchFilter/SearchFilter.vue
```diff
@@ -137,6 +193,13 @@ export default defineComponent({
     SearchDropdowns,
   },
   setup() {
```
AllanJeremy, line 195: Please use `<script setup>` for consistency

AllanJeremy (changes requested):
Couple of changes

AllanJeremy (approved):
:shipit: We'll need to go over some potential cyclic dependencies eg. `filteredLanguage.ts` relies on data from a component.

PR #470, opened 2022-07-18
[angular-ngrx-scss] - Implement type filter functionality

SDaian:
- [x]  when an option is selected, will filter the current results based on the selected option
- [x]  Type dropdown options: all (default), forks, archived
- [x]  only one selection allowed, noted with a check mark icon

Closes: #192


angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
```diff
@@ -7,39 +7,37 @@
...
+      description="Select type"
+      [items]="typeFilters"
+      (setFilter)="handleTypeClick($event)"
+    ></app-repos-filter-dropdown>
+    <app-repos-filter-dropdown
+      name="Sort"
+      description="Select type"
+      [items]="typeFilters"
+      (setFilter)="handleTypeClick($event)"
+    ></app-repos-filter-dropdown>
```
Reshurum, line 28: I assume these are temporary just for styling.

blocker(code-reuse): Right now the styles for the dropdown only live in this ad-hoc component. We will need to re-use them for the other dropdowns in the future as well. I think it would make more sense to create a re-usable dropdown component that lets us pass options in via content projection instead. We could then use that component in all 3 cases here and pass in the relevant filters as children elements.

angular-ngrx-scss/src/app/state/profile/profile.selectors.ts
```diff
@@ -26,25 +26,41 @@ export const selectFilterBySearch = createSelector(
...
     if (search) {
-      return state.repos?.filter((item) =>
+      filteredRepos = filteredRepos?.filter((item) =>
         item.name.toLowerCase().includes(search.toLowerCase()),
       );
     }
-    return state.repos ?? [];
+    if (type !== TypeFilter.All) {
+      if (type === TypeFilter.Archived) {
+        filteredRepos = filteredRepos?.filter((item) => item.archived === true);
```
Reshurum, line 53: suggestion: We can just use the `Boolean` constructor here rather than wrapping things in an arrow function.

```suggestion
        filteredRepos = filteredRepos?.filter(Boolean);
```

angular-ngrx-scss/src/app/state/profile/profile.selectors.ts
```diff
@@ -26,25 +26,41 @@ export const selectFilterBySearch = createSelector(
...
+      }
+      if (type === TypeFilter.Forked) {
+        filteredRepos = filteredRepos?.filter((item) => item.fork === true);
```
Reshurum, line 56: suggestion: We can just use the `Boolean` constructor here rather than wrapping things in an arrow function.

```suggestion
        filteredRepos = filteredRepos?.filter(Boolean);
```

angular-ngrx-scss/src/app/state/profile/profile.state.ts
```diff
@@ -76,6 +76,8 @@ export interface UserReposState {
     url: string;
     node_id: string;
   } | null;
+  fork?: boolean;
+  archived?: boolean;
```
Reshurum, line 80: question: Are these ever set to undefined or null as the result of any API calls?

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-filter-dropdown/repo-filter-dropdown.component.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.isOpen = false;
+  }
+
+  // TODO: maybe convert to directive
+  @HostListener('document:click', ['$event'])
+  onClick(event: PointerEvent) {
+    if (!this.elRef.nativeElement.contains(event.target)) {
+      this.isOpen = false;
+    }
+  }
```
Reshurum, line 73: suggestion: This should be a directive as we can use this functionality elsewhere in a re-usable fashion. I think this ticket would be fine for introducing this behavior. Specifically the navigation bar dropdown could use it.

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
Reshurum, line 28: suggestion: We should make this dropdown a bit more generic, and maybe consider using content projection for the different items instead of passing in an array as an attribute.

Reshurum (changes requested):

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
SDaian: yes, there are just temporary for styling, in the consecutive PRs this will change to the proper name, description, etc.

Agree with you we should re-use a component and passing items via content project looks a great solution, but this is not part of this ticket scope, maybe we need to create another ticket. Just add this component because didn't have anything to complete the functionality described on the AC.
What do you think ? @Reshurum

angular-ngrx-scss/src/app/state/profile/profile.selectors.ts
SDaian: instead of your suggestion added:

`filteredRepos = filteredRepos?.filter((item) => Boolean(item.archived));`

angular-ngrx-scss/src/app/state/profile/profile.selectors.ts
SDaian: `filteredRepos = filteredRepos?.filter((item) => Boolean(item.fork));`

same here.

angular-ngrx-scss/src/app/state/profile/profile.state.ts
SDaian: removed the `?` operator, should always had the value for fork/archived.

angular-ngrx-scss/src/app/home/home.module.ts
```diff
@@ -35,6 +36,7 @@ import { OrganizationInfoComponent } from './profile/org-profile/org-info/org-in
     UserProfileComponent,
     OrgProfileComponent,
     OrganizationInfoComponent,
+    RepoFilterDropdownComponent,
```
asincole, line 39: If you rebase off main branch now, there is a shared module created. Is it better to import and reexport this component there?

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
```diff
@@ -7,39 +7,37 @@
...
   </div>
   <ng-container *ngIf="hasActiveSortAndFilters$ | async">
     <div class="divider"></div>
     <div class="results">
       <p class="results__text">
-        <ng-container *ngIf="selectReposCount$ | async as count">
-          <span class="bold">{{ count }}</span>
+        <ng-container *ngIf="{ data: selectReposCount$ | async } as count">
```
asincole, line 34: [curious]: how does this ```*ngIf="{ data: selectReposCount$ | async } ``` work please?

asincole: or more like, why are we wrapping it in an object first?

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-filter-dropdown/repo-filter-dropdown.component.ts
```diff
@@ -0,0 +1,74 @@
...
+      transition(':enter', [
+        style({ opacity: 0, transform: 'scale(0.95)' }),
+        animate('100ms ease-out', style({ opacity: 1, transform: 'scale(1)' })),
+      ]),
+      transition(':leave', [
+        animate('75ms', style({ opacity: 0, transform: 'scale(0.95)' })),
+      ]),
+    ]),
+  ],
+})
```
asincole, line 31: it seems this component does not mutate itself and gets all its data from a parent? do we mean to change the change detection strategy?

angular-ngrx-scss/src/app/state/profile/profile.reducer.ts
```diff
@@ -14,10 +14,11 @@ const reducer = createReducer(
     ...state,
     ...data,
   })),
-  on(setSortAndFilterProperties, (state, { filters: { search } }) => ({
+  on(setSortAndFilterProperties, (state, { filters: { search, type } }) => ({
     ...state,
     sortAndFilter: {
       search,
+      type,
     },
```
asincole, line 22: [curious] will this action always be called with search and type props? if yes, it's fine, if not, do we want to spread an existing state before updating?

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
SDaian: I'm wrapping in an object because if I rely on the ngIf as it is the `0 results` will not show because 0 is a falsy value.

angular-ngrx-scss/src/app/home/home.module.ts
SDaian: yes, let me rebase and import form there.

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-filter-dropdown/repo-filter-dropdown.component.ts
lindakatcodes: I would agree that we could make the directive, but think I'd also like that in another ticket. Then when it's made, it could also update this and the nav bar to use it. 

- [ ] Please make the ticket to create this directive so we don't forget

angular-ngrx-scss/src/app/state/profile/profile.selectors.ts
```diff
@@ -26,25 +26,41 @@ export const selectFilterBySearch = createSelector(
...
+      }
+    }
+    return filteredRepos ?? [];
   },
 );
 
 export const selectReposCount = createSelector(
   selectRepos,
-  (repos?: UserReposState[]) => repos?.length ?? 0,
+  (repos?: UserReposState[]) => repos?.length,
```
lindakatcodes, line 65: question: Would this still be checking for the length of repos? Would it not need to check the length of `selectRepos` instead? 

Does `repos` always get the full repo list, or does it get updated to be the shorter list if a filter is applied?

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
lindakatcodes: Since this component will be added to the codebase either way, I'd much rather see the reusable component made in this PR and used instead of the temporary one that would then need to be removed and/or recreated later. If we're going to add it either way, let's go ahead and add it the right way.

- [ ] Please go ahead and move this to a shared reusable component

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-filter-dropdown/repo-filter-dropdown.component.ts
SDaian: New issue created.

https://github.com/thisdot/starter.dev-github-showcases/issues/492

angular-ngrx-scss/src/app/state/profile/profile.selectors.ts
SDaian: `repos` is updated after the filter is applied, that's why I create this selector.

angular-ngrx-scss/src/app/state/profile/profile.reducer.ts
SDaian: removed the spread, realize just passing the filters received is Ok to update the sort&Filters property.

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
asincole: Ah, this is interesting lol. In that case, what if we did away with the ng-container entirely?

for example, I am assuming the conditional check if just to hide it till we can get a first value off the observable?
I think (do not know for sure but maybe we can check) we can simply have something like this in that case

```
 <span class="bold">{{ selectReposCount$ | async  }}</span>
```

Directly, I think the number would only get rendered after the first emit, which should achieve the same thing? Let me know what you think about this?

I'm seeing similar direct use in the docs https://angular.io/api/common/AsyncPipe

asincole (approved):

PR #491, opened 2022-07-20
test: completed test, re-wrote test

### LINKED ISSUE #481: [vue3-apollo-quasar] unit test for RepoSubHeader component
# Background
Previous tests were written but didn't follow best practices. We want to update our tests to follow community standards and practices. Some key issues present in our tests include:

- test variables, like the component wrapper, are declared globally in a test `describe` instead of inside of `before*` or `it` scopes
- setup and teardown are done improperly
- conditionals are used in tests rather than using different `describe` scopes or unique test blocks

# Acceptance
- Move variables declarations inside of `before*` or `it` blocks; instantiation can occur inside of `describe` still
- Use setup and teardown properly to establish tests and clean up after them
- migrate conditional tests to either `it.each` blocks OR distinct `describe` blocks with their setup/teardown with the appropriate values set

# Notes 
- Read more about setup and teardown: https://jestjs.io/docs/setup-teardown
- Reducing redundant tests using `it.each`: https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m

hdJerry:
closes #481 
All test re-written


vue3-apollo-quasar/src/components/RepoSubHeader/RepoSubHeader.vue
```diff
@@ -29,8 +29,8 @@
           class="row col col-md-5 col-lg-4 items-center no-wrap justify-between relative-position"
         >
           <button
-            class="git-btn menu__btn text-capitalize row justify-between items-center"
-            @click="toggleWatchMenu"
+            class="git-btn menu_btn--watch text-capitalize row justify-between items-center"
+            @click="toggleWatchMenu()"
```
AllanJeremy, line 33: Context required, what's the convention being used here? BEM? 
Also, why do we have different button type classes: `git-btn` and `menu_btn` with different coding styles?

vue3-apollo-quasar/src/components/RepoSubHeader/RepoSubHeader.vue
```diff
@@ -47,7 +47,10 @@
               <q-icon class="fa fa-caret-down"></q-icon>
             </span>
           </button>
-          <div class="dropdown_menu q-menu" v-if="refWatchMenu">
+          <div
+            class="dropdown_menu q-menu dropdown_menu--watch"
```
AllanJeremy, line 51: Same here, would recommend we pick a convention and stick to it through out the codebase

AllanJeremy (changes requested):
LGTM other than some class naming concerns.
New users of the kit do not have context as to how the decisions were made in terms of CSS.

Therefore, using different conventions within the same codebase without documentation on what the conventions are seems like a slippery slope. 

--
Once again, I am aware that this is a tests PR, but it has brought to light multiple css convention issues that don't scale well. No procrastination please.

vue3-apollo-quasar/src/components/RepoSubHeader/RepoSubHeader.vue
AllanJeremy: @hdJerry

AllanJeremy (approved):

PR #493, opened 2022-07-21
test: completed rewriting test

### LINKED ISSUE #482: [vue3-apollo-quasar] unit test for IssuePullRequestTab component
# Background
Previous tests were written but didn't follow best practices. We want to update our tests to follow community standards and practices. Some key issues present in our tests include:

- test variables, like the component wrapper, are declared globally in a test `describe` instead of inside of `before*` or `it` scopes
- setup and teardown are done improperly
- conditionals are used in tests rather than using different `describe` scopes or unique test blocks

Additionally, only one of the dropdowns in this component were tested and we should test all variants.

# Acceptance
- Move variables declarations inside of `before*` or `it` blocks; instantiation can occur inside of `describe` still
- Use setup and teardown properly to establish tests and clean up after them
- migrate conditional tests to either `it.each` blocks OR distinct `describe` blocks with their setup/teardown with the appropriate values set
- Add tests for missing variants

# Notes 
- Read more about setup and teardown: https://jestjs.io/docs/setup-teardown
- Reducing redundant tests using `it.each`: https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m

hdJerry:
closes #482 
Test rewritten and all test passes

### DIFF
--- a/vue3-apollo-quasar/src/components/IssuePullRequestTab/IssuePullRequestTab.vue
+++ b/vue3-apollo-quasar/src/components/IssuePullRequestTab/IssuePullRequestTab.vue
@@ -3,7 +3,7 @@
     <div class="row items-center no-wrap q-px-xs q-py-sm tab-container">
       <div class="row items-center">
         <button
-          class="tab text-capitalize text-caption q-px-xs row justify-center items-center bg-transparent no-border cursor-pointer"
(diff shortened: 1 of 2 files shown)


AllanJeremy (approved):
:shipit:

PR #494, opened 2022-07-21
chore: route guard added

### LINKED ISSUE #483: [vue3-apollo-quasar] route guard
It was noticed that the Auth component is conditioned in private pages.
what is expected is, if authentication is required for that page user should be routed to the auth page and if possible taken to the route that was visited.
- [x] remove the Auth component conditioned in those pages
- [x] add route guards. 

[Read here for guide about guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#per-route-guard)

hdJerry:
closes #483 

- [x] remove the Auth component conditioned in those pages
- [x] add route guards.


AllanJeremy (approved):
Good PR, good refactor.
LGTM

Issue #195, opened 2022-03-04
[Angular - NgRx - SCSS] Implement search / filter / sort result text


lindakatcodes:
# Implement search / filter / sort result text

## Background

In another ticket, a component was created with various search and filter options. The purpose of this ticket is to dynamically update a line of text that will show below the search / filter component and above the repository list, that gives the user information on how their repositories are filtered and sorted. There should also be a “clear filter” option that shows, that when clicked, removes all filter options and hides this text section.

## Acceptance

- [ ] Below the search bar and above results, should see text related to the current options
- [ ] text will update based on which options are active
    - [ ] search: “{number} results for repositories **matching {query}** sorted by {sort}”
    - [ ] type filter: “{number} results **for {type} repositories** sorted by {sort}”
[ ] language filter:  “{number} results for repositories **written in {language}** sorted by {sort}”
    - [ ] all options: “{number} results for {type} repositories matching {query} written in {language} sorted by {sort}”
    - [ ] sort selection by itself does not show any text - however when any other option is selected, it will show the correct active sort value
- [ ] if search or type / language filter options are selected, will always show “X clear filter” text on the right hand side, which will reset all values to default when clicked

## Notes

PR #496, opened 2022-07-21
[angular-ngrx-scss] - Implement search/filter/sort text

SDaian:
### Description

- [x] Below the search bar and above results, should see text related to the current options
- [x] text will update based on which options are active
    - [x] search: “{number} results for repositories **matching {query}** sorted by {sort}”
    - [x] type filter: “{number} results **for {type} repositories** sorted by {sort}”
    - [x] language filter:  “{number} results for repositories **written in {language}** sorted by {sort}”
    - [x] all options: “{number} results for {type} repositories matching {query} written in {language} sorted by {sort}”
    - [x] sort selection by itself does not show any text - however when any other option is selected, it will show the correct active sort value
- [x] if search or type / language filter options are selected, will always show “X clear filter” text on the right hand side, which will reset all values to default when clicked

image

Closes: #195

### DIFF
--- a/angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
+++ b/angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
@@ -37,6 +37,7 @@
         results
         <ng-container *ngIf="isActiveFilterByType$ | async">
           for <span class="bold">{{ selectFilterByType$ | async }}</span>
+          repositories
         </ng-container>
         <ng-container *ngIf="selectFilterBySearch$ | async as search">
           matching <span class="bold">{{ search }}</span>
--- a/angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
+++ b/angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
@@ -17,8 +17,8 @@
     <app-filter-dropdown
       name="Language"
       description="Select Language"
-      [items]="(filteredLanguages$ | async) || []"
       [current]="selectFilterByLanguage$ | async"
+      [items]="(filteredLanguages$ | async) || []"
       (setFilter)="handleLanguageClick($event)"
     ></app-filter-dropdown>
     <app-filter-dropdown
@@ -52,7 +52,7 @@
           {{ (selectSortFilter$ | async)?.split('_')?.join(' ') | lowercase }}
         </span>
       </p>
-      <span class="results__clear-container">
+      <span class="results__clear-container" (click)="handleClearClick()">
         <div class="close-background">
           <span appOcticon="x" color="white" size="13"></span>
         </div>
--- a/angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
+++ b/angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
@@ -53,12 +53,12 @@
           {{ (selectSortFilter$ | async)?.split('_')?.join(' ') | lowercase }}
         </span>
       </p>
-      <span class="results__clear-container" (click)="handleClearClick()">
-        <div class="close-background">
+      <button type="button" class="results__clear" (click)="handleClearClick()">
+        <div class="results__clear__icon">
           <span appOcticon="x" color="white" size="13"></span>
         </div>
-        <p class="results__clear">Clear filter</p>
-      </span>
+        <p class="results__clear__text">Clear filter</p>
+      </button>
     </div>
     <div class="divider"></div>
   </ng-container>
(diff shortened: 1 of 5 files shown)


lindakatcodes (commented):
Holding on this until a question is answered by the coordinators in the starter-dev Slack channel, as it might require a change. Everything I see so far seems good though!

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.ts
```diff
@@ -135,4 +135,11 @@ export class RepoControlsComponent implements OnInit, OnDestroy {
   handleSortClick(sort: string) {
     this.sortFilter.setValue(sort);
   }
+
+  handleClearClick(): void {
+    this.searchInput.reset();
+    this.typeFilter.setValue(TypeFilter.All);
+    this.languageFilter.setValue(TypeFilter.All);
+    this.sortFilter.setValue(OrderField.UpdatedAt);
```
Reshurum, line 143: blocker(dry): The initial values here don't need to be explicitly passed. The `reset()` method can be used here instead.

```suggestion
    this.typeFilter.reset();
    this.languageFilter.reset();
    this.sortFilter.reset();
```

Reshurum (changes requested):

lindakatcodes (dismissed):

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
```diff
@@ -52,7 +53,7 @@
           {{ (selectSortFilter$ | async)?.split('_')?.join(' ') | lowercase }}
         </span>
       </p>
-      <span class="results__clear-container">
+      <span class="results__clear-container" (click)="handleClearClick()">
```
honzikec, line 56: Could we change `span` to `button type="button"`?

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
```diff
@@ -52,7 +53,7 @@
...
         <div class="close-background">
           <span appOcticon="x" color="white" size="13"></span>
         </div>
```
honzikec, line 59: A `<p>` inside an inline element isn't really semantic, could we change it?

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
```diff
@@ -52,7 +53,7 @@
           {{ (selectSortFilter$ | async)?.split('_')?.join(' ') | lowercase }}
         </span>
       </p>
-      <span class="results__clear-container">
+      <span class="results__clear-container" (click)="handleClearClick()">
         <div class="close-background">
```
honzikec, line 57: Same as below - a div inside inline element is not semantic. Also I would rename the classes to respect BEM convention:

- `results__clear-container` -> `results__clear`
- `results__clear` -> can be left empty or something like `results__clear__text` 
- `close-background` -> `results__clear__icon`

honzikec (commented):
No big issues, but I did have some nitpicks about semantic HTML and class naming convetions 🙂

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
SDaian: changed

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
SDaian: changed

angular-ngrx-scss/src/app/shared/components/repo-controls/repo-controls.component.html
```diff
@@ -52,12 +53,12 @@
           {{ (selectSortFilter$ | async)?.split('_')?.join(' ') | lowercase }}
         </span>
       </p>
-      <span class="results__clear-container">
-        <div class="close-background">
+      <button type="button" class="results__clear" (click)="handleClearClick()">
```
asincole, line 56: [nitpicking]. do we need the type button here? l think that is the default value? Except when setting explicitly to "submit" for example?

honzikec: @asincole, I'm pretty sure the default type is "submit" 🙂

SDaian: Yes indeed, according to w3c the default value for most of the browsers is "Submit". The only browser where the default type is different is in IE.

lindakatcodes (approved):

honzikec (approved):
LGTM!

Issue #202, opened 2022-03-04
[Angular - NgRx - SCSS] Create single repo pull request tab view


lindakatcodes:
# Create single repo pull request tab view

## Background

As a user, if I click into the pull requests tab for a repository, I can see a list of cards containing information about PRs for this repo. By default I land on the view for open PRs, and can click on the words “open” or “closed” to change if I see currently open or closed PRs. 

## Acceptance

- [x] section header shows:
    - [x] {number} open with icon
    - [x] {number} closed with icon
    - [x] label dropdown (does not need options)
    - [x] sort dropdown (does not need options)
- [x] body shows either open or closed (depending on which is selected) pull request cards
- [x] pagination navigation - < prev next > (does not need to function yet)

## Notes

PR #498, opened 2022-07-25
202 create single repo pull request

asincole:
A few things left to be done
* properly build out the view for the pull requests page
* create interface for the pull requests (partial of the existing pull request api response type that only contains props we need)
* update with this new type in necessary places
* write and update test cases as required

closes: #202


honzikec (commented):
No big issues except some TODOs. I can try to help investigate the ActivatedRoute issue further if I have some time.

angular-ngrx-scss/src/app/pull-requests/pull-requests.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  owner = '';
+  repoName = '';
+  pullRequests$ = this.store.select(selectPullRequests);
+
+  constructor(private route: ActivatedRoute, private store: Store) {}
+
+  ngOnInit() {
+    //TODO: for some weird reason, this is not getting the params successfully; investigate
+    this.owner = this.route.snapshot.paramMap.get('owner') as string;
+    this.repoName = this.route.snapshot.paramMap.get('repo') as string;
```
honzikec, line 21:
```suggestion
    const parentRoute = this.route.parent?.parent;
    this.owner = parentRoute?.snapshot.paramMap.get('owner') as string;
    this.repoName = parentRoute?.snapshot.paramMap.get('repo') as string;
```

lindakatcodes (commented):
Just a note on this - I don't know that you're going to need any of the state or service adjustments in here. With Jamie's PR merged, all the pull request data from GitHub and the store should all be here. This should be able to be just the UI aspects of things. Please let me know if that's not true, though!

angular-apollo-tailwind/src/app/pull-requests/pull-requests.module.ts
```diff
@@ -26,6 +26,5 @@ import { IconsModule, PipesModule } from '@shared';
     PaginationModule,
     PipesModule,
   ],
-  exports: [PullRequestsComponent],
```
lindakatcodes, line 29: question: Do we need to be messing with the other repo here? What's the reasoning behind removing this in this PR?

angular-ngrx-scss/src/app/pull-requests/components/pull-requests-header/pull-requests-header.component.html
```diff
@@ -0,0 +1,24 @@
+<div class="filters-container">
+  <div class="pull-requests-status">
+    <a href="" class="active">
+      <span class="icon" appOcticon="no-entry" size="16"></span>
+      <span>412 Open</span>
```
lindakatcodes, line 5: cleanup: Let's make sure the hard coded numbers on lines 5 and 9 are removed before this is merged. :) Should be able to grab those values from the state.

angular-ngrx-scss/src/app/pull-requests/components/pull-requests-header/pull-requests-header.component.html
```diff
@@ -0,0 +1,24 @@
...
+    </a>
+    <a href="">
+      <span class="icon" appOcticon="check" size="16"></span
+      ><span>4877 Open</span></a
```
lindakatcodes, line 9:
```suggestion
      ><span>4877 Closed</span></a
```

angular-ngrx-scss/src/app/pull-requests/components/pull-requests-list/pull-requests-list.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
lindakatcodes, line 1: cleanup: We can delete this test file. In general, ones that are only ensuring the component creates we don't need to keep. :)

angular-ngrx-scss/src/app/pull-requests/pull-requests.component.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
lindakatcodes, line 1: cleanup: Can be removed.

angular-ngrx-scss/src/app/pull-requests/pull-requests.component.ts
```diff
@@ -0,0 +1,29 @@
...
+})
+export class PullRequestsComponent implements OnInit {
+  owner!: string;
+  repoName!: string;
+  pullRequests$ = this.store.select(selectPullRequests);
+
+  constructor(private route: ActivatedRoute, private store: Store) {}
+
+  ngOnInit() {
+    this.owner = this.route.snapshot.paramMap.get('owner') as string;
```
lindakatcodes, line 19: suggestion: Would love to see these values being pulled from the state instead of the route. We should have them there, and selectors for them could be created if the selector doesn't exist yet.

lindakatcodes (changes requested):
There's a few open suggestions from others, and I added a few small things I noticed as well. 

Main suggestion: Updating this PR to use the PR state currently in main instead of creating the new changes in the `state` folder needs to happen. What you had to create here since it didn't exist yet when you started is now in the main branch, so making the updates to use those values instead of adding the new ones would be amazing. :) 

Let me know if there's any questions, would love to help get this merged soon!

angular-ngrx-scss/src/app/pull-requests/pull-requests.component.spec.ts
lindakatcodes: resolving this since more testing was added.

angular-ngrx-scss/src/app/pull-requests/pull-requests.component.ts
lindakatcodes: Can leave this for a future cleanup task, as another task will be working on improving the state structure already.

lindakatcodes (changes requested):
Visually I think we're basically there. There's just a few things we still need to resolve to get this across the finish line.

- I can't tell why, but the comment count for the PR cards appears to be off. If you check it against this repository you can see that some of the PRs don't appear to show the right comment count. Would love to have this investigated, or at the very least a ticket created to solve this later on.
- This PR is a bit out of date and has some merge conflicts. Those will need to be resolved.
- I've gone through and resolved the open comments that were already adjusted or no longer relevant, but there's still a number of comments and changes that have been requested. These need to be addressed, please, even if it's simply to say a reasoning and make a cleanup ticket for it.

angular-apollo-tailwind/src/app/pull-requests/pull-requests.module.ts
asincole: Honestly, can't remember why I did this. will revert

asincole:
> 

@lindakatcodes If I get what you are referencing right (concerning the state), it will not work for me. The existing service methods and interfaces I am seeing align with the PR endpoint GitHub offers, which does not support pagination or getting comments in one call. I had to use the search API to implement this successfully, which led to many bits being different. 

It seems the search API is also returning the wrong comment count or I am missing something because I see what you are saying, but the API response has those values. Most likely, this needs to be investigated. I probably am doing something wrong on that end.

lindakatcodes (approved):
I'm good with this. This gets us a lot of the way there, and we can make a ticket for looking into the comment count issue. Appreciate you getting this finished up! 😁

PR #509, opened 2022-08-02
[remix-gql-tailwind] Feature/repo navigation

### LINKED ISSUE #505: [remix-gql-tailwind] Fix RepoNavigation naming and React Best Practices
# Background
The function [RepoNavigation](https://github.com/thisdot/starter.dev-github-showcases/blob/7c359c21433ec75308d5b4b00bbbccfa270ef5b9/remix-gql-tailwind/app/components/FileExplorerNav/FileExplorerNav.tsx#L7), which somewhat confusingly is the sole and default export from a file named `FileExplorerNav`, is missing React `key`s.

See https://github.com/thisdot/starter.dev-github-showcases/issues/489#issuecomment-1192902473 for original comment

# Acceptance
- Rename FileExplorerNav file that contains RepoNavigation to RepoNavigation
- Update render to include keys in loops

sheerikie:
# What was done?
- Replaced FileExplorerNav to reponavigation
- Added keys to loop item

# Screenshot/Video/GIF of Changes (if applicable)
TODO: @sheerikie 
N/A

# Related Ticket(s)
Resolves #505


remix-gql-tailwind/app/components/RepoNavigation/RepoNavigation.tsx
```diff
@@ -26,7 +26,7 @@ function RepoNavigation() {
             return (
               <>
                 {isLast ? (
-                  <span className={styles.crumbEnd}>{crumb}</span>
+                  <span key={i} className={styles.crumbEnd}>{crumb}</span>
```
dustinsgoodman, line 29: The key needs to be something else other than the index.  I recommend uuid but check out these links from the official docs.

https://beta.reactjs.org/learn/rendering-lists#where-to-get-your-key

image

dustinsgoodman (changes requested):
1 minor but important change request

remix-gql-tailwind/app/components/RepoNavigation/RepoNavigation.tsx
sheerikie: used cuid(already on package.json) for key instead of id

remix-gql-tailwind/app/components/RepoNavigation/RepoNavigation.tsx
```diff
@@ -36,7 +37,7 @@ function RepoNavigation() {
                     >
                       {crumb}
                     </Link>
-                    <span className={styles.separator}>/</span>
+                    <span key={i} className={styles.separator}>/</span>
```
dustinsgoodman, line 40: You need to fix this one too. Generally though, the better key would be to just use the crumbPath or href values.

remix-gql-tailwind/app/components/RepoNavigation/RepoNavigation.tsx
```diff
@@ -26,7 +27,7 @@ function RepoNavigation() {
             return (
               <>
```
tvanantwerp, line 28: Because this fragment is the top-level return from the loop, you'll want to put the `key` here. [Keys are the only props that a React Fragment will accept.](https://reactjs.org/docs/fragments.html#keyed-fragments)

Switching to `<React.Fragment key={whateverTheKeyIs}>...</React.Fragment>` should do it.

Also, I don't think you should use `cuid` here. If I'm understanding the library correctly, each call to `cuid()` generates a new, unique id. Which means on each render, every item will receive a new key. This prevents React from knowing which items experienced no change in order or content, necessitating a full re-render of the entire list and decreasing performance. Using the index `i` has a similar problem, since array contents might change order. I recommend using the `crumb` data to create the key. Assuming `crumb` is a string and that each one is unique, ``key={`crumb-${crumb}`}`` would be a good option.

tvanantwerp (changes requested):
Some changes needed around the insertion of keys for iterating over `crumb`s. Also, why is the `yarn.lock` so different—I don't see any changes to dependencies in `package.json`?

remix-gql-tailwind/app/components/Repositories/RepoNavigation/RepoNavigation.tsx
```diff
@@ -0,0 +1,51 @@
+import { GitBranchIcon } from '../../shared/Icons';
+import { Link } from '@remix-run/react';
+import { useRepo } from '../../../context/RepoContext';
+import * as styles from './RepoNavigation.classNames';
+import { removePathPart } from '../../lib/pathUtils';
+
```
tvanantwerp, line 6: What is this new file for? I don't see it imported anywhere. It's got formatting errors that should've been solved by "format on save" with Prettier, and 3 of the imports are wrong. This looks like it's just a copy of `remix-gql-tailwind/app/components/RepoNavigation/RepoNavigation.tsx`. If there's no purpose to this file, please delete it.

remix-gql-tailwind/app/components/RepoNavigation/RepoNavigation.tsx
```diff
@@ -1,7 +1,7 @@
 import { GitBranchIcon } from '../Icons';
 import { Link } from '@remix-run/react';
 import { useRepo } from '../../context/RepoContext';
-import * as styles from './FileExplorerNav.classNames';
+import * as styles from './RepoNavigation.classNames';
 import { removePathPart } from '../../lib/pathUtils';
```
tvanantwerp, line 6: While things still render fine on the page, it might be worth explicitly importing React. The linter is complaining about it.

image

tvanantwerp (changes requested):

remix-gql-tailwind/app/components/RepoNavigation/RepoNavigation.tsx
sheerikie: thanks..I used the href

remix-gql-tailwind/app/components/RepoNavigation/RepoNavigation.tsx
sheerikie: okay,good...just added it

tvanantwerp (approved):

PR #515, opened 2022-08-05
[angular-ngrx-scss] - Create repo header and navigation view

### LINKED ISSUE #504: [angular-ngrx-scss] add repo header and navigation
# Create single repo view sub header

## Background

As a user, I can view a page for an individual repository with details about that repository. I can see the about information, tabs for code / issues / pull requests, info about how many watchers / stars / forks the repo has, and view the root folder’s contents in a file tree. At the bottom of the page I can also view the contents of the main readme page.

## Acceptance

- [ ] book icon with user name / repo name and visibility tag (public / private)
- [ ] buttons with icon, name, and count for: (buttons do not need to be functional)
    - [ ] watch
    - [ ] star
    - [ ] fork
- [ ] section will have three tab headers for different views (do not need to build the views, just the tab headers)
    - [ ] code w/ < > icon
    - [ ] issues with info icon and count of open issues
    - [ ] pull requests with icon and count of open requests

SDaian:
### Description
- [x] book icon with user name / repo name and visibility tag (public / private)
- [x] buttons with icon, name, and count for: (buttons do not need to be functional)
    - [x] watch
    - [x] star
    - [x] fork
- [x] section will have three tab headers for different views (do not need to build the views, just the tab headers)
    - [x] code w/ < > icon
    - [x] issues with info icon and count of open issues
    - [x] pull requests with icon and count of open requests
    
image

Closes #504

### DIFF
--- a/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
+++ b/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
@@ -1,7 +1,17 @@
 <div class="wrapper">
   <div class="top">
-    <app-repo-heading [owner]="owner" [name]="name"></app-repo-heading>
+    <ng-container *ngIf="repo$ | async as repo">
+      <app-repo-heading
+        [owner]="owner"
+        [name]="name"
+        [repo]="repo"
+      ></app-repo-heading>
+    </ng-container>
     <app-repo-info></app-repo-info>
   </div>
-  <app-repo-navigation [basePath]="basePath"></app-repo-navigation>
+  <app-repo-navigation
+    [basePath]="basePath"
+    [issuesCount]="(repo$ | async)!.issueCount"
+    [pullsCount]="(repo$ | async)!.prCount"
+  ></app-repo-navigation>
 </div>
--- a/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
+++ b/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
@@ -1,11 +1,7 @@
 <div class="wrapper">
   <ng-container *ngIf="repo$ | async as repo">
     <div class="top">
-      <app-repo-heading
-        [owner]="owner"
-        [name]="name"
-        [repo]="repo"
-      ></app-repo-heading>
+      <app-repo-heading [repo]="repo"></app-repo-heading>
       <app-repo-info
         [watch]="repo.watchCount"
         [stars]="repo.starCount"
@@ -13,7 +9,7 @@
       ></app-repo-info>
     </div>
     <app-repo-navigation
-      [basePath]="basePath"
+      [repo]="repo"
       [issuesCount]="repo.issueCount"
       [pullsCount]="repo.prCount"
     ></app-repo-navigation>
--- a/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
+++ b/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
@@ -7,7 +7,11 @@
         [repo]="repo"
       ></app-repo-heading>
     </ng-container>
-    <app-repo-info></app-repo-info>
+    <app-repo-info
+      [watch]="(repo$ | async)!.watchCount"
+      [stars]="(repo$ | async)!.starCount"
+      [forks]="(repo$ | async)!.forkCount"
+    ></app-repo-info>
   </div>
   <app-repo-navigation
     [basePath]="basePath"
--- a/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
+++ b/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
@@ -1,21 +1,21 @@
 <div class="wrapper">
-  <div class="top">
-    <ng-container *ngIf="repo$ | async as repo">
+  <ng-container *ngIf="repo$ | async as repo">
+    <div class="top">
       <app-repo-heading
         [owner]="owner"
         [name]="name"
         [repo]="repo"
       ></app-repo-heading>
-    </ng-container>
-    <app-repo-info
-      [watch]="(repo$ | async)!.watchCount"
-      [stars]="(repo$ | async)!.starCount"
-      [forks]="(repo$ | async)!.forkCount"
-    ></app-repo-info>
-  </div>
-  <app-repo-navigation
-    [basePath]="basePath"
-    [issuesCount]="(repo$ | async)!.issueCount"
-    [pullsCount]="(repo$ | async)!.prCount"
-  ></app-repo-navigation>
+      <app-repo-info
+        [watch]="repo.watchCount"
+        [stars]="repo.starCount"
+        [forks]="repo.forkCount"
+      ></app-repo-info>
+    </div>
+    <app-repo-navigation
+      [basePath]="basePath"
+      [issuesCount]="repo.issueCount"
+      [pullsCount]="repo.prCount"
+    ></app-repo-navigation>
+  </ng-container>
 </div>
--- a/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
+++ b/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
@@ -1,0 +1,7 @@
+<div class="wrapper">
+  <div class="top">
+    <app-repo-heading [owner]="owner" [name]="name"></app-repo-heading>
+    <app-repo-info></app-repo-info>
+  </div>
+  <app-repo-navigation [basePath]="basePath"></app-repo-navigation>
+</div>
--- a/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
+++ b/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
@@ -1,21 +1,21 @@
-import {
-  ChangeDetectionStrategy,
-  Component,
-  Input,
-  OnInit,
-} from '@angular/core';
+import { Component, Input } from '@angular/core';
+import { Store } from '@ngrx/store';
+import { selectedRepository } from 'src/app/state/repository';
 
 @Component({
   selector: 'app-repo-header',
   templateUrl: './repo-header.component.html',
   styleUrls: ['./repo-header.component.scss'],
 })
-export class RepositoryHeaderComponent implements OnInit {
+export class RepositoryHeaderComponent {
   @Input() owner = '';
   @Input() name = '';
-  basePath = '';
 
-  ngOnInit(): void {
-    this.basePath = `/${this.owner}/${this.name}`;
+  repo$ = this.store.select(selectedRepository);
+
+  constructor(private store: Store) {}
+
+  get basePath(): string {
+    return `/${this.owner}/${this.name}`;
   }
 }
--- a/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
+++ b/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
@@ -8,14 +8,7 @@ import { selectedRepository } from 'src/app/state/repository';
   styleUrls: ['./repo-header.component.scss'],
 })
 export class RepositoryHeaderComponent {
-  @Input() owner = '';
-  @Input() name = '';
-
   repo$ = this.store.select(selectedRepository);
 
   constructor(private store: Store) {}
-
-  get basePath(): string {
-    return `/${this.owner}/${this.name}`;
-  }
 }
--- a/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
+++ b/angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
@@ -1,0 +1,21 @@
+import {
+  ChangeDetectionStrategy,
+  Component,
+  Input,
+  OnInit,
+} from '@angular/core';
+
+@Component({
+  selector: 'app-repo-header',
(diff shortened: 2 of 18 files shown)


angular-ngrx-scss/src/app/repository/components/repo-header/repo-info/repo-info.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'app-repo-info',
+  templateUrl: './repo-info.component.html',
+  styleUrls: ['./repo-info.component.scss'],
+})
+export class RepositoryInfoComponent {
+  @Input() watch = 675;
```
lindakatcodes, line 9: suggestion: We should also have these values available, but I'm guessing these are simply default values, correct? If so, I think it'd be best (once you're done testing things) to set these to default to 0.

lindakatcodes (commented):
You might still be working on this, so you might already know about these. But a few things I noticed aren't quite right when viewing the deploy preview: 

- The pull request tab doesn't navigate. Both code and issues work - just not the pull request one.
- On issues, these numbers definitely aren't right. Not sure what's going on, but worth investigating to see where things are going wrong. Maybe they're being added to default counts, or something is getting multiplied somewhere? Way higher than they should be.
- Also on issues, both open and closed counts say "open" after the number. :) 
Screen Shot 2022-08-05 at 5 07 20 PM

angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
```diff
@@ -0,0 +1,21 @@
+import {
+  ChangeDetectionStrategy,
```
honzikec, line 2:
```suggestion
```

angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
```diff
@@ -0,0 +1,21 @@
...
+  templateUrl: './repo-header.component.html',
+  styleUrls: ['./repo-header.component.scss'],
+})
+export class RepositoryHeaderComponent implements OnInit {
+  @Input() owner = '';
+  @Input() name = '';
+  basePath = '';
+
+  ngOnInit(): void {
+    this.basePath = `/${this.owner}/${this.name}`;
```
honzikec, line 19: I wonder whether it may be better to make the basePath a getter instead. That way it would update properly even if we change the owner or name on runtime (I know that'll most likely not happen, but still... 😅 )

Alternative option: make "owner" and "name" into setters and update the basePath there.

angular-ngrx-scss/src/app/repository/components/repo-header/repo-heading/repo-heading.component.html
```diff
@@ -0,0 +1,11 @@
+<div class="heading">
+  <span class="icon" size="24" appOcticon="repo"></span>
+  <div class="heading__container">
+    <a [routerLink]="ownerPath">{{ owner }}</a>
+    <span class="separator">/</span>
```
honzikec, line 5: Nitpick: no css is being set for the .separator class, so it's technically not needed.

angular-ngrx-scss/src/app/repository/components/repo-header/repo-heading/repo-heading.component.html
```diff
@@ -0,0 +1,11 @@
...
+    <a routerLink="./" class="bold">{{ name }}</a>
+  </div>
+  <div class="heading__privacy">
+    <span>Public</span>
```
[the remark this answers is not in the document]
honzikec: Yup, there is a property "_private_". See the repo-card component where it's already being used.

angular-ngrx-scss/src/app/repository/components/repo-header/repo-heading/repo-heading.component.scss
```diff
@@ -0,0 +1,35 @@
+@use '../../../../shared/styles/variables.scss';
+
+.heading {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+
+  &__container {
```
honzikec, line 8: Nitpick: the "heading__container" is such a generic name for an element. I'd prefer something ".heading__path", ".heading__breadcrumbs" or something similar.

angular-ngrx-scss/src/app/repository/components/repo-header/repo-heading/repo-heading.component.scss
```diff
@@ -0,0 +1,35 @@
...
+      text-decoration: none;
+    }
+
+    span {
+      color: variables.$gray600;
+      font-size: 1.475rem;
+    }
+  }
+
+  &__privacy {
```
honzikec, line 26: Let's not forget to account for the privacy states ("public" and "private") when we implement both.

angular-ngrx-scss/src/app/repository/components/repo-header/repo-heading/repo-heading.component.ts
```diff
@@ -0,0 +1,17 @@
...
+  styleUrls: ['./repo-heading.component.scss'],
+})
+export class RepositoryHeadingComponent implements OnInit {
+  @Input() owner = '';
+  @Input() name = '';
+
+  ownerPath = '';
+
+  ngOnInit(): void {
+    this.ownerPath = `/${this.owner}`;
```
honzikec, line 15: Same comment as for the repo-header.component.ts: I'd make this a getter or the owner a setter.

angular-ngrx-scss/src/app/repository/components/repo-header/repo-info/repo-info.component.scss
```diff
@@ -0,0 +1,28 @@
+@use '../../../../shared/styles/variables.scss';
+
+.info-container {
+  display: flex;
+  gap: 0.9rem;
+  .watch,
```
honzikec, line 6: Maybe we can use one class for these buttons as they share the same styles and use the BEM conventions here?

Something like: 

``` SCSS
.info {
  ....
  &__counter {
  ...
    &__icon {
    ...
    }
    &__count {
    ...
    }
  }
}
```

angular-ngrx-scss/src/app/repository/components/repo-header/repo-navigation/repo-navigation.component.scss
```diff
@@ -0,0 +1,63 @@
...
+  margin: 0 auto 0 0;
+  nav {
+    margin-bottom: -1px;
+    display: flex;
+    ul {
+      list-style: none;
+      padding: 0;
+      margin: 0;
+
+      a {
```
honzikec, line 14: Suggestion: use the ".tab" class here and a modifier ".tab--active" / ".tab--inactive" for the states.

angular-ngrx-scss/src/app/repository/components/repo-header/repo-navigation/repo-navigation.component.scss
```diff
@@ -0,0 +1,63 @@
...
+          span {
+            font-weight: 600;
+            color: black;
+
+            &.icon {
+              color: variables.$gray600;
+            }
+          }
+        }
+        &:hover {
```
honzikec, line 33: General comment: When you hover a tab, the repo content jumps 1px down (probably some box-sizing: border-box; needs to be set somewhere). 🙂

angular-ngrx-scss/src/app/repository/components/repo-header/repo-navigation/repo-navigation.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'app-repo-navigation',
+  templateUrl: './repo-navigation.component.html',
+  styleUrls: ['./repo-navigation.component.scss'],
+})
+export class RepositoryNavigationComponent {
+  @Input() basePath = '';
+  @Input() issuesCount = 123;
```
honzikec, line 10: Same comment as Linda left above - probably makes sense to make the default values 0 later.

angular-ngrx-scss/src/app/repository/components/repo-header/repo-heading/repo-heading.component.html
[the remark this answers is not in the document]
SDaian: Added the logic to show private/public info

angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
```diff
@@ -0,0 +1,21 @@
+<div class="wrapper">
+  <div class="top">
+    <ng-container *ngIf="repo$ | async as repo">
```
lindakatcodes, line 3: suggestion: Can this `ng-container` be moved up to wrap around the whole component here, or after the wrapper div? Since all the inner components use the same observable, would make this a bit cleaner and require less observable checking.

angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
```diff
@@ -0,0 +1,21 @@
...
+import { selectedRepository } from 'src/app/state/repository';
+
+@Component({
+  selector: 'app-repo-header',
+  templateUrl: './repo-header.component.html',
+  styleUrls: ['./repo-header.component.scss'],
+})
+export class RepositoryHeaderComponent {
+  @Input() owner = '';
+  @Input() name = '';
```
lindakatcodes, line 12: question: Do we need the owner and name coming in as inputs here? We're getting the repo from the store, which should have this information - can we not use those values instead?

angular-ngrx-scss/src/app/repository/components/repo-header/repo-heading/repo-heading.component.ts
```diff
@@ -0,0 +1,26 @@
...
+  @Input() owner = '';
+  @Input() name = '';
+
+  get ownerPath(): string {
+    return `/${this.owner}`;
+  }
+
+  get visibility(): string {
+    if (this.repo) {
+      return this.repo.visibility ? 'Private' : 'Public';
```
lindakatcodes, line 21: suggestion: Visibility should be a string and not a boolean in the store - this getter should return the visibility value directly. Maybe something like:
```suggestion
      return this.repo.visibility || '';
```
So it'll default back to empty if that value isn't set, but otherwise returns the value?

angular-ngrx-scss/src/app/repository/components/repo-header/repo-navigation/repo-navigation.component.scss
```diff
@@ -0,0 +1,68 @@
+@use '../../../../shared/styles/variables.scss';
```
lindakatcodes, line 1: issue: Leaving this here since it's CSS related - I don't think our active tab highlight is working _quite_ like we'd expect. In the deploy preview, I'm seeing some odd alternation on hover, and when clicking on the issues route, the highlight stays under the code tab instead of toggling off. 

Not sure exactly what's causing it, but worth looking into! 

https://user-images.githubusercontent.com/16214572/184669590-6132e05b-c852-4ad6-aa28-98cc12f59bbe.mov

angular-ngrx-scss/src/app/repository/components/repository-details/repository-details.component.ts
```diff
@@ -1,8 +1,19 @@
...
   styleUrls: ['./repository-details.component.scss'],
 })
-export class RepositoryDetailsComponent {}
+export class RepositoryDetailsComponent implements OnInit {
+  owner = '';
+  repo = '';
+
+  constructor(private route: ActivatedRoute) {}
+
+  ngOnInit(): void {
```
lindakatcodes, line 15: suggestion: Would really prefer these values be pulled in from the store. This info already exists there. Creating selectors for them if they don't already exist would be wonderful.

lindakatcodes (changes requested):
Left a few suggestions and a CSS issue with the active tab highlight, but this feels really close!

angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.html
SDaian: moved to the div after the wrapper, you were right they are using the same observable

angular-ngrx-scss/src/app/repository/components/repo-header/repo-header.component.ts
SDaian: they are coming from different places, but there is a reason why. 
We are accessing that data through the route, and passing it to the repo-header via Input(), because we already had that data available, but the repo-header info needs to be accessible via the info in the Store.

That makes sense to you ?

angular-ngrx-scss/src/app/repository/components/repository-details/repository-details.component.ts
SDaian: think IMHO, waiting for the store to retrieve the owner and the name is not worth it, coming from the org view when you already have the owner and using the info from the `paramMap` without waiting to resolve an API is better that waiting for the selector, because at a certain point when we move to this URL for ex `getsentry/sentry` we don't have the data.

Maybe we can bring up this talk to the shared channel :)

angular-ngrx-scss/src/app/repository/components/repo-header/repo-navigation/repo-navigation.component.scss
SDaian: solved.

lindakatcodes (approved):
This looks great!! Appreciate your receptiveness to feedback. ❤️ 

There's one small unused input that would be good to clean up in the `repo-header`, but otherwise this is good to go!

Issue #489, opened 2022-07-19
[remix-gql-tailwind] Architectural review of current kit & showcase and any cleanup recommendations


dustinsgoodman:

tvanantwerp:
## README

The `README.md` file contains only a minimum amount of information about Storybook and authenticating with GitHub. The majority of the file hasn’t been edited in 6 months. Doesn’t explain many things, e.g., Arc.

## Storybook

Storybook is not officially supported by Remix yet. Since this starter’s initial implementation of Storybook, there’s been a great deal of detailed discussion about how Remix should support it [here](https://github.com/remix-run/remix/discussions/2481).

## `.env.example`

Does not include all env variables required. Missing `GITHUB_CLIENT_ID`, `GITHUB_CLIENT_SECRET`, and `BASE_URL`.

## Running the project

Could not even get the project to run. After adding necessary `.env` variables and clicking the Login with GitHub button, I ended up on an error screen. The login button correctly goes to GitHub, then returns to the correct `/auth/github/callback` route with the necessary codes. I used the same GitHub app credentials in `.env` as when I started work on the project six months ago. On redirect to the root `/` route, the server returns a 500 error. The CatchBoundary in `root.tsx` is used, but rendered as text rather than HTML. CLI states: **Error: GraphQL Error (Code: 405)**.

Full error:

```
Error

[...]

PR #516, opened 2022-08-05
[remix-gql-tailwind] Updates from audit notes

### LINKED ISSUE #506: [remix-gql-tailwind] Update README with newly relevant information
# Background

The `README.md` file contains only a minimum amount of information about Storybook and authenticating with GitHub. The majority of the file hasn’t been edited in 6 months. Doesn’t explain many things, e.g., Arc.

## Storybook

Storybook is not officially supported by Remix yet. Since this starter’s initial implementation of Storybook, there’s been a great deal of detailed discussion about how Remix should support it [here](https://github.com/remix-run/remix/discussions/2481).

## `.env.example`

Does not include all env variables required. Missing `GITHUB_CLIENT_ID`, `GITHUB_CLIENT_SECRET`, `GITHUB_GRAPHQL_ENDPOINT`, and `BASE_URL`. The `GITHUB_GRAPHQL_ENDPOINT` in particular isn't likely to vary and may benefit from being included as a default value. Presently, the GraphQL client initialization file falls back to a localhost URL which doesn't do anything, causing errors.

## Running the project

~Could not even get the project to run. After adding necessary `.env` variables and clicking the Login with GitHub button, I ended up on an error screen. The login button correctly goes to GitHub, then returns to the correct `/auth/github/callback` route with the necessary codes. I used the same GitHub app credentials in `.env` as when I started work on the project six months ago. On redirect to the root `/` route, the server returns a 500 error. The CatchBoundary in `root.tsx` is used, but rendered as text rather than HTML. CLI states: **Error: GraphQL Error (Code: 405)**.~

The app didn't display correctly due to a missing `GITHUB_GRAPHQL_ENDPOINT` env variable, which was undocumented.

Storybook didn’t work properly, either. The mocks to replace <Link> with <a> didn’t seem to be working as expected, and many components displayed errors as a result.

```
useHref() may be used only in the context of a <Router> component.
```

# Acceptance
- Update the readme per the above instructions

tvanantwerp:
Changes based on notes in #489. Resolves #506.

- `.env.example` updated with all known variables
- Storybook documentation updated
- Broken Storybook mocks fixed
- Updates to testing with your own GitHub OAuth app
- Remove auto-generated `index.js.map` file from version control


remix-gql-tailwind/.env.example
```diff
@@ -1,2 +1,6 @@
 NODE_ENV="development"
 SESSION_SECRET="super-duper-s3cret"
+GITHUB_CLIENT_ID=
+GITHUB_CLIENT_SECRET=
```
dustinsgoodman, line 4: A couple notes:
1. If we put these in GitHub secrets for whatever reason, they need different names as these are reserved in GitHub actions.
2. For a valid secrets file, we should put at least `xxx` on the right hand of the equals

dustinsgoodman (approved):

PR #527, opened 2022-08-20
[Angular - NgRx - SCSS] 440: state & service refactor

### LINKED ISSUE #440: [Angular - NgRx - SCSS] Refactor of Services, Prep for state restructuring
**Update Sept. 16:**

We've decided to split this work into multiple tickets, to make it easier to track and review. The work this ticket is tracking is updated below, with new tickets created for the remaining work needed.

This ticket should have:
- [x] Services adjusted to only call the API and return the full response
- [x] Store folder structure verified, and any new folders / files created so that there is a folder for each page of the application
- [x] The site should still "work" - a user can still navigate to any part of the application that has currently been completed
- [x] The auth folder in the store will be updated to include a small piece of information related to the authenticated user, so that information is always available and can be used for other API calls

-----------------

The structure of the NgRx Store State is getting a little convoluted and could potentially cause issues as we keep working on this showcase. 

A few things to work on that will help with this:
- [ ] Store files should relate to the app page, and inner files should be relocated to where they'd go in the adjusted structure
- [x] The services might be doing too much data manipulation. A pass would be good to ensure all services are simply calling the API and returning the API response, which either the effect or reducer can then massage into the shape and names we want.

Related conversation for reference:
suggestion: I think these should be attached to home page specific state rather than being a part of the global user state; however I know we discussed possibly refactoring some of the state management in this showcase, so we can elect to fix that later if that makes more sense.

_Originally posted by @Reshurum in https://github.com/thisdot/starter.dev-github-showcases/pull/319#discussion_r912121472_

lindakatcodes:
Resolves #440


lindakatcodes:
This part of the refactor is focused on:
- adjusting services to return the full data from the API
- adjusting state files to shape the data into what we expect to get back
- initial setup of the new dashboard state files
- any small changes needed to ensure the app still works and functions exactly how it should after the cleanup changes

This does still need to be rebased, but will have to hold for awhile since the files that are conflicting will need some thorough review to match the point of this refactor.

lindakatcodes:
This has now been rebased and is as updated as I can get it - the site still works as expected. 

There is still work to do with the repository services - some things got duplicated and it's a bit too messy even with this refactor to completely unify everything, so there's a bit of overlap between some of the service calls. I unified them as best as I could, but future work would be helpful here.

Also I got most of the test files adjusted, but I've still got 4 failing tests:
- UserEffects: the user gists test doesn't equal what it should
- UserEffects: get user info from API has an undefined error, I'm guessing from setting up the mock store but no mock data so this could (hopefully) be easily fixed
- RepositoryService: info on provided repository doesn't equal what it should
- UserService: user data from API doesn't equal what it should

The 3 tests that don't seem to match visually look like they do, so perhaps it's something minor somewhere that's mixed up. Needs a bit more checking to be sure.

angular-ngrx-scss/src/app/app.module.ts
```diff
@@ -11,6 +11,7 @@ import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
 import { TokenInterceptor } from './auth/services/token.interceptor';
 import { reducers } from './state';
+// import { DashboardEffects } from './state/dashboard/dashboard.effects';
```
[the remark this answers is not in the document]
lindakatcodes: This comment is intentional and I'd like to leave it as part of the refactor effort. This piece of state has been set up but the data hasn't been migrated to it yet. Would be preferable to simply uncomment this when it's ready.

angular-ngrx-scss/src/app/app.module.ts
```diff
@@ -27,7 +28,12 @@ import { UserEffects } from './state/user';
       logOnly: environment.production,
       autoPause: true,
     }),
-    EffectsModule.forRoot([UserEffects, ProfileEffects, RepositoryEffects]),
+    EffectsModule.forRoot([
+      // DashboardEffects,
```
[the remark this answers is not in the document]
lindakatcodes: This comment is intentional and I'd like to leave it as part of the refactor effort. This piece of state has been set up but the data hasn't been migrated to it yet. Would be preferable to simply uncomment this when it's ready.

angular-ngrx-scss/src/app/state/auth/auth.effects.ts
```diff
@@ -49,9 +71,62 @@ export class AuthEffects {
...
+  //           };
+  //           return fetchAuthenticatedUserDataSuccess({ userData: user });
+  //         }),
+  //         catchError((error) =>
+  //           of(fetchAuthenticatedUserDataFailure({ error })),
+  //         ),
+  //       );
+  //     }),
+  //   );
+  // });
```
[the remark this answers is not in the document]
lindakatcodes: Already removed once I found the fix for this. 😁

angular-ngrx-scss/src/app/repository/services/repository.service.ts
```diff
@@ -93,106 +66,88 @@ export class RepositoryService {
...
-          })),
-        };
-      }),
-    );
-  }
+  ): Observable<PullRequestAPIResponse> {
+    const owner = encodeURIComponent(repoOwner);
+    const name = encodeURIComponent(repoName);
+    const state = encodeURIComponent(prState);
+    const url = `${environment.githubUrl}/search/issues?q=repo:${owner}/${name}+type:pr+state:${state}`;
```
ktrz, line 133: an interesting idea just popped into my head to avoid having to explicitly encode all the params. We could use a tagged template to achieve the same thing with less boilerplate.
Just putting it as a discussion point and it definitely can be done on a separate ticket/PR

angular-ngrx-scss/src/app/state/auth/auth.effects.ts
```diff
@@ -49,9 +70,40 @@ export class AuthEffects {
...
+  /**
+   * Gets authenticated user's name, photo, and email
+   */
+  fetchAuthUserData$ = createEffect(() => {
+    return this.actions$.pipe(
+      ofType(saveUserTokenSuccess, userTokenExists),
+      switchMap(() =>
+        this.store.select(selectAuthUserName).pipe(
+          skipWhile((name) => name.length > 0),
+          mergeMap(() =>
```
ktrz, line 82: do we actually want to use `mergeMap` here? If for some reason the selector emits multiple times we would end up with multiple requests happening at once and a race condition. Maybe this is the rare place to use `exhaustMap`?

angular-ngrx-scss/src/app/state/auth/auth.effects.ts
```diff
@@ -49,9 +70,40 @@ export class AuthEffects {
...
 
+  /**
+   * Gets authenticated user's name, photo, and email
+   */
+  fetchAuthUserData$ = createEffect(() => {
+    return this.actions$.pipe(
+      ofType(saveUserTokenSuccess, userTokenExists),
+      switchMap(() =>
+        this.store.select(selectAuthUserName).pipe(
+          skipWhile((name) => name.length > 0),
```
ktrz, line 81: just curious. What is the intention here for using `skipWhile`?

angular-ngrx-scss/src/app/state/auth/auth.selectors.ts
```diff
@@ -9,3 +9,13 @@ export const selectIsAuthenticated = createSelector(
...
+
+export const selectAuthUserAvatar = createSelector(
+  selectAuthState,
+  (state: AuthState) => state.authUser.avatar,
+);
+
+export const selectAuthUserName = createSelector(
+  selectAuthState,
+  (state: AuthState) => state.authUser.username,
+);
```
ktrz, line 21: I suggest using more selectors composition. This way you reuse some of the logic as well as can get some performance benefits. Additionally you don't need to explicitly provide types for the `state` etc. `createSelector` is strictly typed and all those types will be inferred

```suggestion

const selectAuthUser = createSelector(selectAuthState, state => state.authState)

export const selectAuthUserAvatar = createSelector(
  selectAuthUser,
  authUser => authUser.avatar,
);

export const selectAuthUserName = createSelector(
  selectAuthUser,
  authUser => authUser.username,
);
```

angular-ngrx-scss/src/app/state/dashboard/dashboard.reducer.ts
```diff
@@ -0,0 +1,15 @@
+import { Action, createReducer } from '@ngrx/store';
+import { DashboardState } from './dashboard.state';
+
+export const initialState: DashboardState = {
+  temp: '',
```
ktrz, line 5: let's not pollute the state with unnecessary properties

angular-ngrx-scss/src/app/state/profile/profile.effects.ts
```diff
@@ -16,12 +18,54 @@ export class ProfileEffects {
...
+              name: userData.name,
+              twitter_username: userData.twitter_username,
+              username: userData.login,
+              type: userData.type,
+            };
+            const orgs: UserOrgsState[] = orgsData.map((org) => ({
+              id: org.id,
+              login: org.login,
+              avatar_url: org.avatar_url,
+            }));
```
ktrz, line 43: I think it would make this code much more readable if we extracted the mapping functions instead of inlining them here

angular-ngrx-scss/src/app/state/profile/profile.effects.ts
```diff
@@ -16,12 +18,54 @@ export class ProfileEffects {
...
+              avatar: userData.avatar_url,
+              bio: userData.bio,
+              blog: userData.blog,
+              company: userData.company,
+              email: userData.email,
+              followers: userData.followers,
+              following: userData.following,
+              location: userData.location,
+              name: userData.name,
+              twitter_username: userData.twitter_username,
```
ktrz, line 35: IMO we shouldn't keep the fields' names exactly as they are in the API response but rather convert to conventions that we use in JS/TS ie. use camelCase instead of snake_case

angular-ngrx-scss/src/app/state/profile/profile.state.ts
```diff
@@ -86,26 +86,7 @@ export interface UserRepo {
...
-    starred_url: string;
-    subscriptions_url: string;
-    organizations_url: string;
-    repos_url: string;
-    events_url: string;
-    received_events_url: string;
-    type: string;
-    site_admin: boolean;
-  };
+  owner: UserApiResponse;
```
ktrz, line 89: I don't think we should store the information in the state in the exact shape as data returned from API. Or at least keep 2 separate interfaces for API and State data.

angular-ngrx-scss/src/app/state/repository/repository.selectors.ts
```diff
@@ -1,26 +1,26 @@
...
-export const repositoryFeatureKey = 'repo';
+export const repositoryFeatureKey = 'repository';
 export const selectRepositoryState =
-  createFeatureSelector<RepoState>(repositoryFeatureKey);
+  createFeatureSelector<RepositoryState>(repositoryFeatureKey);
 
 export const selectedRepository = createSelector(
   selectRepositoryState,
-  (state: RepoState) => state,
+  (state: RepositoryState) => state,
```
ktrz, line 10: we don't need to explicitly type the `state` parameter. `createSelector` is strictly typed and will infer that correctly from the return type of `selectRepositoryState`

angular-ngrx-scss/src/app/state/repository/repository.selectors.ts
```diff
@@ -1,26 +1,26 @@
...
+import { RepositoryState } from './repository.state';
 
-export const repositoryFeatureKey = 'repo';
+export const repositoryFeatureKey = 'repository';
 export const selectRepositoryState =
-  createFeatureSelector<RepoState>(repositoryFeatureKey);
+  createFeatureSelector<RepositoryState>(repositoryFeatureKey);
 
 export const selectedRepository = createSelector(
   selectRepositoryState,
```
ktrz, line 9: why do we even have this selector if it's not actually selecting anything different than the whole state?

angular-ngrx-scss/src/app/state/repository/repository.selectors.ts
```diff
@@ -1,26 +1,26 @@
...
-  (state: RepoState) => state,
+  (state: RepositoryState) => state,
 );
 
 export const selectCurrentlySelectedFile = createSelector(
   selectRepositoryState,
-  (state: RepoState) => state.selectedFile,
+  (state: RepositoryState) => state.selectedFile,
```
ktrz, line 15: we don't need to explicitly type the `state` parameter. `createSelector` is strictly typed and will infer that correctly from the return type of `selectRepositoryState`

angular-ngrx-scss/src/app/state/app.reducer.ts
```diff
@@ -1,13 +1,15 @@
 import { ActionReducerMap } from '@ngrx/store';
 import { AppState } from './app.state';
 import { authReducer } from './auth/auth.reducer';
+// import { dashboardReducer } from './dashboard/dashboard.reducer';
```
ktrz, line 4: can we remove the commented out code?

angular-ngrx-scss/src/app/user/services/user.service.ts
```diff
@@ -19,153 +15,109 @@ import { environment } from 'src/environments/environment';
...
-        twitter_username: data.twitter_username,
-        username: data.login,
-        type: data.type,
-      })),
-    );
+    return this.http.get<UserApiResponse>(url, {
+      headers: {
+        Accept: 'application/vnd.github.v3+json',
+      },
+    });
```
ktrz, line 30: same as in other service. Can we move this to an interceptor?

angular-ngrx-scss/src/app/app.module.ts
```diff
@@ -27,7 +29,13 @@ import { UserEffects } from './state/user';
...
+      AuthEffects,
+      UserEffects,
+      ProfileEffects,
+      RepositoryEffects,
+    ]),
```
ktrz, line 38: is splitting the reducers out of the global declaration into a forFeature setup part of this PR?

ktrz (changes requested):

angular-ngrx-scss/src/app/app.module.ts
lindakatcodes: No, it's not. Everything right now is intended to be global.

angular-ngrx-scss/src/app/file-viewer/file-explorer-blob/file-explorer-blob.component.spec.ts
```diff
@@ -15,14 +15,14 @@ describe('FileExplorerBlobComponent', () => {
...
-    repo: {
+    repository: {
       selectedFile: {
         content: 'this is a readme file',
         name: 'starter.dev-github-showcases',
         type: 'file',
         size: 223,
       },
-    } as RepoState,
+    } as RepositoryState,
```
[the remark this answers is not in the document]
lindakatcodes: We are. The `RepositoryState` contains information for the current repository, and this test appears to be written specifically to test an individual file. So it does the casting to focus directly on the `selectedFile` that it's testing. 

The `AppState` also appears to be needed, as the tests fail without it. I'm guessing because they're global it needs to be wrapped inside to be read properly? But I'm not certain on that.

angular-ngrx-scss/src/app/home/home.component.ts
```diff
@@ -9,11 +9,13 @@ import { fetchUserData } from '../state/user/user.actions';
...
-  user$ = this.store.select(selectUserLoginName);
+  user$ = this.store.select(selectAuthUserName);
 
   constructor(private store: Store) {}
 
   ngOnInit() {
-    this.store.dispatch(fetchUserData());
+    this.user$.subscribe((user) => {
+      this.store.dispatch(fetchUserData({ username: user }));
+    });
```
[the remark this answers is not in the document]
lindakatcodes: Oh good catch. The subscription is only here because we need that name from the `user$` selector and this was the simplest way I could figure to pass it to the store effect it's already dispatching. 

Updated this to destroy the call when it's done.

angular-ngrx-scss/src/app/pull-requests/pull-requests.component.spec.ts
```diff
@@ -6,21 +6,21 @@ import { MockStore, provideMockStore } from '@ngrx/store/testing';
...
-    repo: {
+    repository: {
       selectedFile: {
         content: 'this is a readme file',
         name: 'starter.dev-github-showcases',
         type: 'file',
         size: 223,
       },
-    } as RepoState,
+    } as RepositoryState,
```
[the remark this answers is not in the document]
lindakatcodes: Same as the previous one - the writer of this test was focusing on a particular slice of the state and didn't want to write out the entire `RepositoryState` object.

angular-ngrx-scss/src/app/repository/services/repository.service.ts
```diff
@@ -93,106 +66,88 @@ export class RepositoryService {
...
-          // TODO: consider using a function that also takes encoding format to decode this
-          content: atob(data.content),
-          size: data.size,
-        };
-      }),
-    );
+    return this.http.get<IssueComments>(url, {
+      headers: {
+        Accept: 'application/vnd.github.v3+json',
+      },
```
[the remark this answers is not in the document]
lindakatcodes: This is a smart idea. Created a ticket to make this interceptor in a future PR: https://github.com/thisdot/starter.dev-github-showcases/issues/711

angular-ngrx-scss/src/app/repository/services/repository.service.ts
lindakatcodes: What do you mean by tagged template? Would be curious to see an idea of what that might look like. Could then create an improvement ticket for that!

angular-ngrx-scss/src/app/state/app.reducer.ts
```diff
@@ -1,13 +1,15 @@
...
 import { userReducer } from './user/user.reducer';
-import { repoReducer } from './repository/repository.reducer';
 
 export const reducers: ActionReducerMap<AppState> = {
   auth: authReducer,
-  user: userReducer,
+  // dashboard: dashboardReducer,
   profile: profileReducer,
-  repo: repoReducer,
+  repository: repositoryReducer,
```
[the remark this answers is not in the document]
lindakatcodes: Answered more fully in a previous comment, but no that's not part of this PR.

angular-ngrx-scss/src/app/state/auth/auth.selectors.ts
lindakatcodes: Appreciate this!

angular-ngrx-scss/src/app/state/profile/profile.effects.ts
lindakatcodes: The `twitter_username` was pretty easy to adjust, but a lot of the others are tied more directly to API calls. We have a ticket created already to adjust the interfaces - I'm going to mark in there to also try to adjust these a bit more for the app-facing ones so we're consistent.

angular-ngrx-scss/src/app/state/profile/profile.state.ts
lindakatcodes: We have a ticket created already to clean up and improve the interfaces. I'll add the state files to be checked and cleaned to that as well.

angular-ngrx-scss/src/app/state/repository/repository.selectors.ts
lindakatcodes: Valid point. It's used in two different component files, but I agree that it doesn't really seem super necessary. I'll make a note to check into this whenever this section of the store is refactored.

angular-ngrx-scss/src/app/state/user/user.effects.ts
```diff
@@ -12,16 +21,38 @@ import {
...
+                  email: userData.email,
+                  followers: userData.followers,
+                  following: userData.following,
+                  location: userData.location,
+                  name: userData.name,
+                  twitter_username: userData.twitter_username,
+                  username: userData.login,
+                  type: userData.type,
+                };
+                return fetchUserDataSuccess({ userData: user });
```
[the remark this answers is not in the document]
lindakatcodes: Sorry Chris, I'm not sure I understand. Do you mean we should create some functions in the state files that would handle mapping from API response to app response, and then pull in those functions and use them in the effects? Instead of directly doing the mapping here?

angular-ngrx-scss/src/app/user/services/user.service.ts
lindakatcodes: Acknowledged in the other comment - created a ticket to do this in a separate PR.

angular-ngrx-scss/src/app/repository/services/repository.service.spec.ts
```diff
@@ -257,25 +231,60 @@ describe('RepositoryService', () => {
...
+          expect(httpClientSpy.get).toHaveBeenCalledWith(
+            `https://api.github.com/repos/thisdot/starter.dev-github-showcases`,
+            jasmine.objectContaining({
+              headers: {
+                Accept: 'application/vnd.github.v3+json',
+              },
+            }),
+          );
+        },
+        complete: done,
```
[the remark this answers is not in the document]
lindakatcodes: For now I've adjusted the call inside `next` to be `toHaveBeenCalledOnceWith`, so it checks it was only called once. Hopefully that should catch this state?

angular-ngrx-scss/src/app/state/auth/auth.effects.ts
ktrz: wouldn't it make more sense to use `distinctUntilChanged` then?

angular-ngrx-scss/src/app/state/repository/repository.selectors.ts
ktrz: can we put a TODO comment in here to not forget about this?

angular-ngrx-scss/src/app/repository/services/repository.service.ts
ktrz: have you created a ticket for that?

angular-ngrx-scss/src/app/state/profile/profile.effects.ts
ktrz: have you created a ticket for using mapping functions? AFAIR we discussed putting a POC in this PR and then following up for other places in another PR

ktrz (commented):
Looks great! Just a couple of housekeeping comments to make sure we have everything that we've discussed captured in tickets

angular-ngrx-scss/src/app/repository/services/repository.service.ts
lindakatcodes: Ticket here: https://github.com/thisdot/starter.dev-github-showcases/issues/824

angular-ngrx-scss/src/app/state/auth/auth.effects.ts
lindakatcodes: Oops missed that one, this is fixed now.

angular-ngrx-scss/src/app/state/profile/profile.effects.ts
lindakatcodes: Ticket: https://github.com/thisdot/starter.dev-github-showcases/issues/825

angular-ngrx-scss/src/app/state/repository/repository.selectors.ts
lindakatcodes: Added!

ktrz (approved):
🚢 🚢 🚢

Issue #591, opened 2022-09-12
[Angular-Apollo-Tailwind] Fix relative links to the app root


Atrophius:
Absence of  `http://` in profile causes links to render as relative to the app root. 

### Problem: 
![Image](https://user-images.githubusercontent.com/39612839/189439440-1e4659dc-1359-4b1d-9d7b-62fec30af4c0.png)

### Fix example:

![Image](https://user-images.githubusercontent.com/39612839/189439481-71d04b23-9ff6-4cbf-b58f-59090e9d01f0.png)

PR #637, opened 2022-09-16
fix(angular-apollo-tailwind): append protocol if missing from user url

asincole:
Refs: #591

### DIFF
--- a/angular-apollo-tailwind/projects/shared/src/lib/pipes/generate-url-with-protocol/generate-url-with-protocol.pipe.spec.ts
+++ b/angular-apollo-tailwind/projects/shared/src/lib/pipes/generate-url-with-protocol/generate-url-with-protocol.pipe.spec.ts
@@ -1,0 +1,8 @@
+import { GenerateUrlWithProtocolPipe } from './generate-url-with-protocol.pipe';
+
+describe('GenerateUrlWithProtocolPipe', () => {
+  it('create an instance', () => {
+    const pipe = new GenerateUrlWithProtocolPipe();
+    expect(pipe).toBeTruthy();
+  });
+});
--- a/angular-apollo-tailwind/projects/shared/src/lib/pipes/generate-url-with-protocol/generate-url-with-protocol.pipe.spec.ts
+++ b/angular-apollo-tailwind/projects/shared/src/lib/pipes/generate-url-with-protocol/generate-url-with-protocol.pipe.spec.ts
@@ -5,4 +5,19 @@ describe('GenerateUrlWithProtocolPipe', () => {
     const pipe = new GenerateUrlWithProtocolPipe();
     expect(pipe).toBeTruthy();
   });
+
+  it('should append https to url with no protocol', () => {
+    const pipe = new GenerateUrlWithProtocolPipe();
+    expect(pipe.transform('test.com')).toBe('https://test.com');
+  });
+
+  it('should return url untouched if it has protocol appended', () => {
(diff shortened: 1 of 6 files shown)


angular-apollo-tailwind/src/app/profile/profile-about/profile-about.component.html
```diff
@@ -74,7 +74,7 @@ <h1 class="mt-2">
           ></hero-icon>
           <a
             class="link"
-            [href]="user.websiteUrl"
+            [href]="generateValidUrl(user.websiteUrl)"
```
ktrz, line 77: Could you please implement this as pipe? Calling functions in template is not good for performance

ktrz (dismissed):

angular-apollo-tailwind/src/app/profile/profile-about/profile-about.component.html
asincole: oh sure.

asincole: done @ktrz

angular-apollo-tailwind/src/app/profile/profile-about/profile-about.component.html
```diff
@@ -88,7 +88,7 @@ <h1 class="mt-2">
           ></sd-twitter-icon>
           <a
             class="link"
-            [href]="'https:/twitter.com/' + user.twitterUsername"
+            [href]="'https://twitter.com/' + user.twitterUsername"
```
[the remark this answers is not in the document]
asincole: there is already a check that only displays this if its not null or undefined a few lines above it. as to your suggestion 

@ktrz what do you think about this bit?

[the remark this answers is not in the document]
asincole: @ihardz look at this comment above https://github.com/thisdot/starter.dev-github-showcases/pull/637#discussion_r972992991 on this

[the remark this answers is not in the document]
ktrz: Moving this logic out of the component is definitely a great idea but we shouldn't call functions from a component in the template for performance reasons so pipe seems like a better place. This way you could also add some basic unit tests easily.

Feel free to do it as part of this PR or just create a new issue + PR

angular-apollo-tailwind/projects/shared/src/lib/pipes/generate-url-with-protocol/generate-url-with-protocol.pipe.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { GenerateUrlWithProtocolPipe } from './generate-url-with-protocol.pipe';
+
+describe('GenerateUrlWithProtocolPipe', () => {
+  it('create an instance', () => {
+    const pipe = new GenerateUrlWithProtocolPipe();
+    expect(pipe).toBeTruthy();
+  });
```
ktrz, line 7: Let's add couple of basic unit tests for that logic :)
Unit tests should be part of each new PR. Especially if we're fixing a bug

lindakatcodes (commented):
Think this looks reasonable to me, but would be good to have Chris or Ihar re-review it as well! I didn't have a ton of time left this week to thoroughly test this, but it seems like everything is in place.

lindakatcodes (approved):
Tested again locally and appears to be working as expected!

PR #647, opened 2022-09-22
chore: updated repository card to fit ticket criterias

### LINKED ISSUE #631: [nuxt-pinia-tailwind] Create repository card
## Background

This will be a reusable component that displays details about a repository. On the home page and profile page, a user will see a list of their repositories. This component is intended to be the building block for that - a single card for a repository, showing pertinent data about that repo.

## Acceptance

- [ ] Card view for use on repository lists
- [ ] should always have:
    - [ ] repo name
    - [ ] visibility tag (public / private)
    - [ ] description (if available)
    - [ ] main language with color dot
    - [ ] star icon with number of stars on repo
    - [ ] last updated text
    - [ ] optional - star button on right to add a star to the repo (does not need to function yet)

## Notes

temitopeadesoji:
Closes #631


nuxt-pinia-tailwind/plugins/formatter.ts
```diff
@@ -0,0 +1,50 @@
...
+      addSuffix?: boolean;
+      locale?: Locale;
+    }
+  ) => {
+    if (typeof date === 'string') {
+      date = new Date(date);
+      return formatDistance(date, dateToCompare, options);
+    }
+    if (isValidDate(date) && isValidDate(dateToCompare)) {
+      return formatDistance(date, dateToCompare, options);
```
jesus4497, line 40: nit: I think we could reduce this to just one condition to make it easier to read
```suggestion
      if (!isValidDate(date) || !isValidDate(dateToCompare)) {
        return ''
      }
      const dateToFormat = typeof date === 'string' ? date : new Date(date)
      return formatDistance(date, dateToCompare, options); 
      
```

jesus4497 (changes requested):
LGTM! just a small change

hdJerry (approved):
LGTM

nuxt-pinia-tailwind/components/RepositoryCard/RepositoryCard.vue
```diff
@@ -41,6 +47,16 @@ export default {
...
     },
+    meta: {
+      type: Object as PropType<{
+        language: string;
+        updatedAt: Date;
+        forkCount: number;
+        stargazerCount: number;
+      }>,
+      required: true,
+      default: () => {},
```
jesus4497, line 58: This is actually returning `void` not an actual object.
```suggestion
      default: () => ({
       language: '',
       updatedAt: new Date(),
       forkCount: 0,
       stargazerCount: 0
      }),
```

jesus4497 (changes requested):
Just a small change! We should also add tests for this components

hdJerry (approved):

PR #650, opened 2022-09-22
feat: create user profile card

temitopeadesoji:


nuxt-pinia-tailwind/schemes/authScheme.ts
```diff
@@ -40,21 +40,28 @@ export default class CustomLoginScheme extends LocalScheme {
...
       //
 
-      const [{ data: orgs }, { data: stars }] = await Promise.all([
-        getOrgs,
-        getStars,
-      ]);
+      const [orgs, stars] = await Promise.all([getOrgs, getStars]);
+
+      const starsCount =
+        stars.headers?.link?.split(',')[1].match(/.*page=(?<page_num>\d+)/)
```
hdJerry, line 55: What is we made this a re-useable function. This function can be where other helper functions are if there be any.

```ts
const pageCounter = (value: string) => value?.split(',')[1].match(/.*page=(?<page_num>\d+)/)?.groups?.page_num || 0;

const starsCount = pageCounter(stars.headers?.link);
const orgsCount = pageCounter(orgs.headers?.link);
```

nuxt-pinia-tailwind/components/TabNavigation/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface TabItem {
```
jesus4497, line 1: nit: could we rename this to `ITabItem` instead? Just to keep following the naming convention of interfaces in the project

nuxt-pinia-tailwind/components/UserProfile/UserProfile.vue
```diff
@@ -0,0 +1,92 @@
+<template>
+  <div>
+    <div class="w-[260px] h-[260px]">
+      <img
+        :src="userProfile.avatar"
+        alt="Avatar"
+        width="260"
+        height="260"
+        class="absolute rounded-full shadow z-30"
+      />
```
jesus4497, line 10: I was having an error here in my IDE, following [this](https://github.com/nuxt/typescript/issues/537#issuecomment-1050054649) helped me solve the issue. Just installing the package was enough for me, I didn't have to change the tsconfig file

image

nuxt-pinia-tailwind/components/UserProfile/UserProfile.vue
```diff
@@ -0,0 +1,92 @@
...
+      return {
+        avatar: this.$auth.user.avatar_url,
+        name: this.$auth.user.name,
+        username: this.$auth.user.login,
+        bio: this.$auth.user.bio,
+        followers: this.$auth.user.followers,
+        following: this.$auth.user.following,
+        company: this.$auth.user.company,
+        location: this.$auth.user.location,
+        websiteUrl: this.$auth.user.website_url,
```
jesus4497, line 86: We should update our types here
image

nuxt-pinia-tailwind/schemes/authScheme.ts
jesus4497: This is nice! It helps to make the call quicker instead of just calling the whole thing. I agree with @hdJerry, we should make this a utility function though

jesus4497 (changes requested):

jesus4497:
The link of the screenshot is broken :(

hdJerry:
Screenshot 2023-08-01 at 15 21 58

hdJerry (approved):

Megio (approved):

PR #662, opened 2022-09-30
feat: create main page layout

markshenouda:
image

image

### DIFF
--- a/nuxt-pinia-tailwind/components/UserRepos/UserRepos.vue
+++ b/nuxt-pinia-tailwind/components/UserRepos/UserRepos.vue
@@ -23,7 +23,8 @@
 <script lang="ts">
 import { useAsync, useContext } from '@nuxtjs/composition-api';
 import Vue from 'vue';
-import RepositoryCard from '~/components/RepositoryCard/RepositoryCard.vue;
+// import RepositoryCard from '~/components/RepositoryCard/RepositoryCard.vue;
+import RepositoryCard from '~/components/RepositoryCard/RepositoryCard.vue';
 import { useUserStore } from '~/store/userStore';
 
 export default Vue.extend({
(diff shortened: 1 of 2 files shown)


nuxt-pinia-tailwind/components/UserRepos/UserRepos.vue
```diff
@@ -1,13 +1,44 @@
...
+import RepositoryCard from '../RepositoryCard/RepositoryCard.vue';
+import { useUserStore } from '~/store/userStore';
+
+export default Vue.extend({
+  name: 'UserRepos',
+  components: { RepositoryCard },
+  setup() {
+    const userStore = useUserStore();
+    const { $auth } = useContext();
+    useAsync(async () => {
```
jesus4497, line 35: In order to avoid the client-side rendered virtual DOM tree not matching the server-rendered content, we need to listen to the data returned by `useAsync`

```suggestion
    const repos = useAsync(async () => {
      await userStore.getUserRepos();

      return userStore.repos
    });

    return {
      repos,
      username: $auth.$state.user.login,
    };
```

jesus4497 (changes requested):

nuxt-pinia-tailwind/components/UserRepos/UserRepos.vue
```diff
@@ -1,13 +1,46 @@
...
+      >
+    </div>
   </div>
 </template>
 
 <script lang="ts">
-export default {};
+import { useAsync, useContext } from '@nuxtjs/composition-api';
+import Vue from 'vue';
+import RepositoryCard from '../RepositoryCard/RepositoryCard.vue';
```
jesus4497, line 26: Lets do an alias here, instead of `../` lets  do: 

```suggestion
import RepositoryCard from '~/components/RepositoryCard/RepositoryCard.vue;
```

jesus4497 (changes requested):

PR #663, opened 2022-10-12
chore: Update issue templates to align with Dev expectation

Zelig880:


.github/custom-issue-templates/01 Initialize showcase with KIT.md
```diff
@@ -0,0 +1,12 @@
+# Initialize showcase with KIT
+
+## Background
+
+As a developer I would like my starting base to be the existing KIT for the current tehcnology. It is expected for a KIT to be fully ready before commencing work on the showcase.
```
dustinsgoodman, line 5:
```suggestion
As a developer, I would like my starting base to be the existing KIT for the current technology. It is expected for a KIT to be fully ready before commencing work on the showcase.
```

dustinsgoodman (commented):
I don't like deferring the app deployment so far because seeing it in the wild is part of the main deliverable. Can we change that back?

Zelig880:
@dustinsgoodman I have moved it back, but I think it would be good to be a Milestone 2 deliverable, also because for some of this new tech, deployment can be an issue and I was trying to avoid it to be on day one! :)

dustinsgoodman (approved):

PR #725, opened 2022-10-26
feat(svelte-kit-scss): create top repos component 

asincole:
Closes: #674


svelte-kit-scss/src/routes/home/+page.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
```
flakolefluk, line 1: is this file intended to be here?

svelte-kit-scss/src/routes/(home)/+page.server.ts
```diff
@@ -0,0 +1,26 @@
...
+        // TODO: how do I get the token
+        Authorization: `Bearer `
+      }
+    })
+    const data = (await response.json()) as UserReposApiResponse;
+    return {
+      topRepos: (mapUserReposToTopRepos(data))
+    }
+  } catch (err) {
+    console.error(err)
```
flakolefluk, line 23: i think you should also either throw an error or return  an empty object, that should make `data` always defined in the page

svelte-kit-scss/src/lib/helpers/repositories.ts
```diff
@@ -0,0 +1,32 @@
+import type {UserReposApiResponse} from "../interfaces";
+import type {UserReposState} from "../interfaces";
```
flakolefluk, line 2: can you make it a single import?

iansamz (approved):
LGTM

PR #729, opened 2022-10-26
chore: updated amplify.yml file

hdJerry:
# Setup app deployment for new project

## Acceptance

- [ ] AWS setup for new showcase with working url
- [ ] Adjust the production environment file to include the new deployment url

## Notes

- We use AWS Amplify for our deployments
- If you do not have access to the AWS account, please ask a manager for access
- You should have to modify the `amplify.yaml` in the top-level directory

### DIFF
--- a/amplify.yml
+++ b/amplify.yml
@@ -135,11 +135,11 @@ applications:
         preBuild:
           commands:
             - nvm install --lts=gallium
-            - pnpm install
+            - npm install
         build:
           commands:
             - node -v
-            - pnpm run build
+            - npm run build
       artifacts:
         baseDirectory: dist/
         files:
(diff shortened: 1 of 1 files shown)


amplify.yml
```diff
@@ -125,6 +125,25 @@ applications:
...
+        paths:
+          - node_modules/**/*
+
+  - appRoot: solidjs-tailwind
+    frontend:
+      phases:
+        preBuild:
+          commands:
+            - nvm install --lts=gallium
+            - pnpm install
```
mbicknese, line 138: Not sure if nvm supplies `pnpm` by default. Do we need to install it?

kodejuice: I think we need to install it, no first class support yet

vyktoremario: @hdJerry what's the update on this?

hdJerry: We can use nvm then.

vyktoremario (approved):

PR #745, opened 2022-10-28
chore: workflow added

hdJerry:
# Setup CI for new project

## Background

We want to review changes in as near real-time as possible. The goal of this is to set up guard rails for development practices and validate features as they’re completed.

## Acceptance

- [x] Setup linting
- [x] Setup Unit testing
- [x] Setup build

## Notes
You can use existing [CI workflows](https://github.com/thisdot/starter.dev-github-showcases/tree/main/.github/workflows) as a baseline

### DIFF
--- a/.github/workflows/ci-solidjs-tailwind.yml
+++ b/.github/workflows/ci-solidjs-tailwind.yml
@@ -2,11 +2,15 @@ name: Solidjs TailwindCSS CI
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - "solidjs-tailwind/**"
     paths:
       - "solidjs-tailwind/**"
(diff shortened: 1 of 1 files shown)


mbicknese:
Oh wait, it only runs on the main branch! the `path` refers to which files it needs to look at. @hdJerry can you check what happens if we remove the `branches` filter?

vyktoremario:
@hdJerry where are we on this?

hdJerry:
I will revert soon

hdJerry:
I think we have to merge to main to see it it works for other branches, and if it doesn't we can create a ticket to fix.
@vyktoremario @mbicknese

vyktoremario (approved):

PR #791, opened 2022-10-31
[qwik-pod] - Get Organization repositories from GitHub API

hdJerry:
# Get Organization repositories from GitHub API

## Background

There will be a view for an organization’s profile, which, similar to the user’s profile, will list all of the organization’s repositories.

## Acceptance

- [x] Get organization repositories from GitHub API
- [x] For display on organization page; data needed:
    - [x] Repo name
    - [x] description
    - [x] language
    - [x] star count
    - [x] fork count
    - [x] last updated
    - [x] visibility tag

## Notes

- This query should start sorted by most recently updated.


AllStackDev1 (changes requested):
The api should be in the organization’s profile page

PR #810, opened 2022-11-02
[qwik-pod]:c Implement type filter functionality

hdJerry:
# Implement type filter functionality

## Background

In another ticket, a component was created with the type dropdown button. The purpose of this ticket is to add the functionality and populate the fields for that button. The options that should be available to the user are listed below - when a user selects an option, the repository list on the page should be filtered to only include ones that match the user’s selection.

## Acceptance

## Notes

Screenshot 2022-11-02 at 15 39 27
Screenshot 2022-11-02 at 15 39 38


SDaian (approved):
LGTM

PR #820, opened 2022-11-03
[qwik-pod]:  Implement language filter functionality

hdJerry:
# Implement language filter functionality

## Background

In another ticket, a component was created with the language dropdown button. The purpose of this ticket is to add the functionality and populate the fields for that button. The options that should be available to the user are listed below - when a user selects an option, the repository list on the page should be filtered to only include ones that match the user’s selection.

## Acceptance

- [x] language options should show all (default) and list of languages from users’ repos

## Notes

Screenshot 2022-11-03 at 12 32 41
Screenshot 2022-11-03 at 12 32 48

### DIFF
--- a/qwik-graphql-tailwind/src/components/user-repos/user-repos.tsx
+++ b/qwik-graphql-tailwind/src/components/user-repos/user-repos.tsx
@@ -9,12 +9,15 @@ import { Pagination } from '../pagination/pagination';
 import { RepoFilters } from '../repo-filters/repo-filters';
 import { getLanguages } from './getLanguages';
 import filterStore from '~/context/repo-filter';
-import { repoDataFilteredBySearch } from './filter-sort-functions';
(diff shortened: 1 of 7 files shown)


SDaian (approved):
LGTM

qwik-graphql-tailwind/src/components/user-repos/user-repos.tsx
```diff
@@ -24,7 +27,7 @@ export const UserRepos = component$(({ repos, owner }: UserReposProps) => {
   return (
     <>
       <RepoFilters languages={languages} resultCount={repos.nodes.length} />
-      {filteredAndSortedRepos.map(
+      {languageFilter.map(
```
SDaian, line 31: why we are iterating over the languageFilter array, we should use `filteredAndSortedRepos`, please make that change

SDaian (changes requested):

Issue #691, opened 2022-10-20
[solidjs-pod] 15 Create main page layout (top repositories + gists)


Zelig880:
# Create main page layout

## Background

As a user, my main page view should show me my gists and top repositories.

## Acceptance

 - [ ] 2 column layout
 - [ ] Left side
    - [ ] gists panel
 - [ ] Right side
    - [ ] Top repositories title
    - [ ] card for each of the users’ top repos
    - [ ] view all repositories button that navigates to user profile page

PR #844, opened 2022-11-08
feat(solidjs): Create main layout(gists + top repos)

SDaian:
## Acceptance Criteria

- [x]  2 column layout
      - [ ]  Left side
       - [ ]  gists panel
- [x]  Right side
      - [ ]  Top repositories title
      - [ ]  card for each of the users’ top repos
- [x]  view all repositories button that navigates to user profile page

Closes: #691

## Preview

image


vyktoremario (commented):
There is a ticket for creating the `RepoCard` component. 
I think this PR went out of scope.

Also, about the style, did we decide to place them in a different file?

SDaian:
> There is a ticket for creating the `RepoCard` component. I think this PR went out of scope.
> 
> Also, about the style, did we decide to place them in a different file?

yes, makes sense @vyktoremario, think in the moment I started to work on this, the other ticket wasn't merged yet, that's why. Will remove that. Thanks!! :)

solidjs-tailwind/src/components/PrivacyBadge/PrivacyBadge.classNames.js
```diff
@@ -0,0 +1 @@
+export const badge = 'py-0.5 px-2 text-xs rounded-xl text-gray-600 border border-gray-300 font-medium';
```
vyktoremario, line 1: Instead of using this method of styling, we should use a CSS module just the wayit was implemented in Nextjs.

This pattern was used in Remix because Remix does not support CSS module.

vyktoremario (changes requested):
I left a comment.

Also, I saw that some `index.js` files were deleted, is there a reason for this?

AllStackDev1 (changes requested):
There is a lot of merge conflict and some files where refactored and deleted without reason.

danecando:
> I left a comment.
> 
> Also, I saw that some `index.js` files were deleted, is there a reason for this?

In most cases an index file that re-exports the entire directory is not needed. Generally imports should from the file directly. I think one of the main exceptions to this is a directory that is a single component but includes files for other things like tests etc.

solidjs-tailwind/src/components/PrivacyBadge/PrivacyBadge.classNames.js
danecando: I agree I don't love this pattern either I'll update these

danecando (dismissed):

AllStackDev1 (approved):
It looks ok, but i could see we added back the Octokit which was removed so we stick with fetch gql. You should be open to issue that might arise from this merge.

PR #862, opened 2022-11-08
chore(svelte-kit-scss): fix failing tests

asincole:
This fixes some tests that were broken or did not work right. Also, I noticed that using "$lib/**/**" for imports seems to be breaking tests, this switches back to relative parts. I have not investigated deeply enough so there may be a fix that lets us use that type of import when testing, else, we want to make note to use relative paths when importing stuff.

Closes: #44


iansamz:
I have found how to fix this issue by adding $lib alias on vitest.config.ts

```
import { defineConfig } from 'vitest/config';
import { svelte } from '@sveltejs/vite-plugin-svelte';
import path from 'path';

export default defineConfig({
  plugins: [svelte({ hot: !process.env.VITEST })],
  test: {
    globals: true,
    environment: 'jsdom',
  },
  resolve: {
    alias: {
      $lib: path.resolve(__dirname, './src/lib'),
    },
  },
});

```

PR #870, opened 2022-11-11
[sveltekit-pod] 33 Create file view

### LINKED ISSUE #861: [sveltekit-pod] 33 Create file view
# Create file view

## Background

As a user, if I’m viewing a single repository and click on an individual file, I am navigated to a single page view of that file. This should look similar to how a file would look on GitHub itself.

## Acceptance

- [ ] clicking on a file in a repo displays that file

## Notes

ihardz:
resolves #861

> # Create file view
> 
> ## Background
> 
> As a user, if I’m viewing a single repository and click on an individual file, I am navigated to a single page view of that file. This should look similar to how a file would look on GitHub itself.
> 
> ## Acceptance
> 
> - [x] clicking on a file in a repo displays that file
> 
> ## Notes


iansamz (approved):
Awesome work. I believe its good to ship

PR #871, opened 2022-11-11
feat(qwik): PR tab view

SDaian:
## Acceptance Criteria

- [x]  section header shows:
     - [x]  {number} open with icon
     - [x]  {number} closed with icon
     - [x]  label dropdown (does not need options)
     - [x]  sort dropdown (does not need options)
- [x]  body shows either open or closed (depending on which is selected) pull request cards
- [x]  pagination navigation - < prev next > (does not need to function yet)

## Preview

image

Closes: #847


hdJerry (changes requested):
This ticket seems to have gone out of the scope. please look into it. something doesn't sync.
@kenslachtajr please help clarify

SDaian:
> This ticket seems to have gone out of the scope. please look into it. something doesn't sync. @kenslachtajr please help clarify

sorry, @hdJerry don't know which is the "out of scope" thing, please instead of just leaving that comment, clarify which part is out of scope. The only thing I did differently from yours has added the routing, which makes sense because, in order to see my page and therefore yours, we need the routing working, deliver a thing we can only test via Storybook is not my taste.

cc: @kenslachtajr

AllStackDev1 (approved):
This PR might have a conflict with mine depends on which is merged first... will keep an eye out.

LGTM

PR #875, opened 2022-11-14
feat: [sveltekit-pod] Show file views for selected folders

### LINKED ISSUE #876: [sveltekit-pod] Show file views for selected folders
# Background
As a user
I want to see a list of nested files and folders.
So that I can click on folder on the repository files view and navigate to the folder files view. The same for nested folders.

# Acceptance Criteria:
- [ ] Add appropriate routing
- [ ] Reuse existing components whenever possible.\

ihardz:
resolves #876

> # Background
> As a user
> I want to see a list of nested files and folders.
> So that I can click on folder on the repository files view and navigate to the folder files view. The same for nested folders.
> 
> # Acceptance Criteria:
> - [x] Add appropriate routing
> - [x] Reuse existing components whenever possible.
>

### DIFF
--- a/svelte-kit-scss/src/lib/components/Repo/RepoHeader/RepoHeader.svelte
+++ b/svelte-kit-scss/src/lib/components/Repo/RepoHeader/RepoHeader.svelte
@@ -5,7 +5,7 @@
   import RepoNavigation from './RepoNavigation/RepoNavigation.svelte';
 
   export let repo: RepoState;
-  const { watchCount, starCount, forkCount, openIssuesCount, openPullRequestsCount } = repo;
+  $: ({ watchCount, starCount, forkCount, openIssuesCount, openPullRequestsCount } = repo);
 </script>
 
 <div class="wrapper">


svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/(file-view)/[...tree]/+page.svelte
```diff
@@ -3,17 +3,18 @@
...
   import FileExplorerReadme from '$lib/components/FileExplorer/FileExplorerReadme/FileExplorerReadme.svelte';
-  import type { LayoutServerData } from './$types';
+  import type { PageServerData } from './$types';
 
-  export let data: LayoutServerData;
-  const { repoInfo, username, repo } = data;
-  const { description, website, tags, readme, tree, activeBranch } = repoInfo;
+  export let data: PageServerData;
+  console.log('DATA', data);
+  const { repoInfo, username, repo, readme, tree } = data;
```
Comment (BigAB) [inline svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/(file-view)/[...tree]/+page.svelte:10]: This is a bit of a "svelteism" but if you destructure an object (or array) and you need the properties to be reactive you need to use the "reactive destructuring" syntax [described in the svelte docs](https://svelte.dev/docs#component-format-script-3-$-marks-a-statement-as-reactive), which is basically, _in a reactive statement destructure as an expression_.

```suggestion
  $: ({ repoInfo, username, repo, readme, tree } = data);
```

...once you change them to this you should see the components updating correctly. It's a tricky little thing but now that you know it it'll only bite you one or two more times.

We probably want to look for other instances of destructuring in svelte components and change them to this style too.

svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/(file-view)/[...tree]/+page.svelte
```diff
@@ -3,17 +3,18 @@
...
+  const { description, website, tags, defaultBranch } = repoInfo;
```
Comment (BigAB) [inline svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/(file-view)/[...tree]/+page.svelte:11]:
```suggestion
  $: ({ description, website, tags, defaultBranch } = repoInfo);
```

svelte-kit-scss/src/lib/components/Repo/RepoHeader/RepoHeader.svelte
```diff
@@ -5,7 +5,7 @@
   import RepoNavigation from './RepoNavigation/RepoNavigation.svelte';
 
   export let repo: RepoState;
-  const { watchCount, starCount, forkCount, issueCount, prCount } = repo;
+  const { watchCount, starCount, forkCount, openIssuesCount, openPullRequestsCount } = repo;
```
BigAB, line 8:
```suggestion
  $: ({ watchCount, starCount, forkCount, openIssuesCount, openPullRequestsCount } = repo);
```

svelte-kit-scss/src/lib/components/FileExplorer/FileExplorerAbout/FileExplorerAbout.svelte
```diff
@@ -1,10 +1,10 @@
 <script lang="ts">
   import { Book16, Link16 } from 'svelte-octicons';
   import { handleAnchorClick } from '$lib/helpers';
+  import type { RepoState } from '$lib/interfaces';
 
-  export let description: string;
-  export let website: string;
-  export let tags: string[];
+  export let repoInfo: RepoState;
+  const { description, website, tags } = repoInfo;
```
BigAB, line 7: In svelte components, destructuring like this will not be "reactive", so when, in this case if `description`, `website` or tags `tags` were to change, they would not update the elements and/components they are passed to in the template.

So in svelte components, you should (almost) always destructure using the Reactive Statements syntax, as [described in the svelte docs](https://svelte.dev/docs#component-format-script-3-$-marks-a-statement-as-reactive):
```js
$: ({ name } = person)
```

Otherwise it will surprise someone else someday, when things are not updating correctly.
This is very much only a svelte thing, but it is a very common svelte thing, and we should definitely follow the Svelte conventions for `starter.dev` projects

svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/(file-view)/[...tree]/+page.svelte
```diff
@@ -3,25 +3,27 @@
...
+  import type { PageServerData } from './$types';
+  export let data: PageServerData;
 </script>
 
 <div class="container grid grid-cols-12 subpage">
   <section class="col-span-9 col-sm-span-12">
     <FileExplorerNav />
-    <FileExplorerContainer contents={tree} branch={activeBranch} {username} {repo} />
+    <FileExplorerContainer
+      folder={data?.folder}
```
Comment (BigAB) [inline svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/(file-view)/[...tree]/+page.svelte:14]: None of these properties should be `undefined` or `null` (except `data.userInfo` which isn't used in this template) so there should be no need for the optional chaining operator `?.` for any of these

svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.server.ts
```diff
@@ -1,29 +1,50 @@
...
-  const getRepoContentsUrl = new URL(`/repos/${username}/${repo}/contents`, ENV.GITHUB_URL);
-  const getRepoReadmeUrl = new URL(`/repos/${username}/${repo}/readme`, ENV.GITHUB_URL);
 
-  const [repoData, repoPRs, repoContentsData, repoReadmeData] = await Promise.all([
+  const getOpenRepoPullsUrl = new URL(`/search/issues`, ENV.GITHUB_URL);
+  getOpenRepoPullsUrl.searchParams.append('q', `repo:${username}/${repo} is:pr is:open`);
+  getOpenRepoPullsUrl.searchParams.append('per_page', '1');
+
+  // const getRepoContentsUrl = new URL(`/repos/${username}/${repo}/contents`, ENV.GITHUB_URL);
+  // const getRepoReadmeUrl = new URL(`/repos/${username}/${repo}/readme`, ENV.GITHUB_URL);
```
Comment (BigAB) [inline svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.server.ts:20]: Is this a WIP, do these comments need to be removed, or are you leaving them in because they will be uncommented in a later PR?

svelte-kit-scss/src/lib/components/FileExplorer/models.ts
```diff
@@ -0,0 +1,12 @@
+import type { GithubRepoContentsItemType } from '$lib/constants/github';
```
BigAB, line 1: Honest question, how are we distinguishing `models` vs `interfaces`?

 - `src/lib/components/FileExplorer/models.ts`
 - `src/lib/interfaces`

BigAB (changes requested):
The question about the commented out code pertains to all the commented out code in this PR?
(I'm okay with leaving them in, but only if there is supposed to be a follow-up PR to uncomment them, otherwise perhaps they should be removed?)

I am requesting the change for using the svelte Reactive statements style of destructuring in svelte components because, this is a starter.dev showcase, so we should use it as an opportunity to inform svelte users that this is the way to do it in svelte components

svelte-kit-scss/src/lib/components/FileExplorer/models.ts
BigAB: Personally I like the idea of splitting it into 3 things, each with their own interfaces:

 - **getting the data** (from an API and verifying with zod or something for types)
 - **the data** (or state or store) which is the types the front-end application "has" or expects
 - **showing the data** (the components) which optimally DON'T use the types from the store/state, they just have their own "props" interfaces and they base everything off of that
 
 ...and none of these 3 layers (...or systems, or modules or whatever to call them) share types
 
 **THAT BEING SAID**: I think for these demo apps, the 3 split might be a bit much, but we could go for just a split of:
 
- interfaces used by functions/classes (Services?) for getting the data 
- mapping functions from 1 type to another
- components having their own type (the models here)

So maybe we should put the types in `src/lib/interfaces` in with the things that fetch data?
(This may be a bit of a rabbit hole/sidetrack that doesn't need to be resolved this PR or maybe not even in this project)

BigAB (approved):
LGTM if it's out of the WIP stage
:rocket:

PR #877, opened 2022-11-15
Fix Error occuring on build

iansamz:
Unexpected token error when building the project with some commonjs modules

if (typeof global !== 'undefined') { global.Prism = Prism; }

https://thisdot-labs.slack.com/archives/C0475BB41DW/p1668521420727819


kenslachtajr (approved):
LGTM

PR #881, opened 2022-11-15
[qwik-pod]: Mocking API in storybook

hdJerry:
The purpose here is to set up the mock API (MSW or StoryBook add-on). The goal is to be able to mock the API in the components that are fetching actual data.

Screenshot 2022-11-15 at 21 30 15
Screenshot 2022-11-15 at 21 30 28
Screenshot 2022-11-15 at 21 30 36


SDaian (approved):
LGTM

PR #917, opened 2022-11-19
Update repo header component use cases

AllStackDev1:
### Acceptance criteria 

- [x] Update the other use cases
- [x] Update the links to pass appropriate data when navigating to pages that require this data.

### DIFF
--- a/qwik-graphql-tailwind/src/components/repo-header/index.tsx
+++ b/qwik-graphql-tailwind/src/components/repo-header/index.tsx
@@ -11,8 +11,8 @@ interface RepoHeaderProps {
   stargazerCount: number | string;
   forkCount: number | string;
   watcherCount: number | string;
-  issuesCount: number;
-  prCount: number;
+  issuesCount: number | string;
+  prCount: number | string;
 }
 
 export const RepoHeader = component$(
@@ -20,8 +20,8 @@ export const RepoHeader = component$(
     const { pathname } = useLocation();
     const basePath = `${owner}/${name}`;
     const tabList = createTabList({
-      issueCount: issuesCount,
-      pullRequestCount: prCount,
+      issueCount: Number(issuesCount),
+      pullRequestCount: Number(prCount),
     });
 
     return (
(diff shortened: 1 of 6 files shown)


qwik-graphql-tailwind/src/components/repo-header/index.tsx
```diff
@@ -11,17 +11,17 @@ interface RepoHeaderProps {
   stargazerCount: number | string;
   forkCount: number | string;
   watcherCount: number | string;
-  issuesCount: number;
-  prCount: number;
+  issuesCount: number | string;
+  prCount: number | string;
```
hdJerry, line 15: why did you update this by adding string?
I think it's only appropriate that the counts are numbers. Please reverse.

qwik-graphql-tailwind/src/components/repo-header/index.tsx
```diff
@@ -11,17 +11,17 @@ interface RepoHeaderProps {
...
 
 export const RepoHeader = component$(
   ({ name, owner, watcherCount, stargazerCount, forkCount, issuesCount, prCount }: RepoHeaderProps) => {
     const { pathname } = useLocation();
     const basePath = `${owner}/${name}`;
     const tabList = createTabList({
-      issueCount: issuesCount,
-      pullRequestCount: prCount,
+      issueCount: Number(issuesCount),
+      pullRequestCount: Number(prCount),
```
hdJerry, line 24: why do we have to convert to number? worst case scenario is that it would be Zero(0)

hdJerry (changes requested):

SDaian (approved):
LGTM

qwik-graphql-tailwind/src/components/repo-header/index.tsx
AllStackDev1: Because we are also, sending it as query params which means they will be strings

qwik-graphql-tailwind/src/components/repo-header/index.tsx
AllStackDev1: The issue and PR count will be used for calculations(addition) and they made be a string, which will result in unwanted concatenation.

AllStackDev1: Note that, because we can't pass values around via context we made use of query params.

qwik-graphql-tailwind/src/components/repo-header/index.tsx
hdJerry: so if you convert to Number, then why does this have number and string?

AllStackDev1: It is been used in a function that requires it to be a Number.
image

PR #920, opened 2022-11-21
Qwik graphql tailwind/mvp

AllStackDev1:


.gitignore
```diff
@@ -9,7 +9,7 @@
 
 # IDE - VSCode
 .vscode/*
-!.vscode/settings.json
+.vscode/settings.json
```
Zelig880, line 12: Why are we making this change? is it necessary? (this will be applied to all showcases and I do not think it to be needed.

.gitignore
```diff
@@ -18,3 +18,5 @@
 # System Files
 .DS_Store
 node_modules
+
+.vscode
```
Zelig880, line 22: same as above

Zelig880 (changes requested):
Question about gitignore changes

.gitignore
AllStackDev1: Hmm, I will comment it out.

.gitignore
AllStackDev1: Will remove this

Zelig880 (approved):

PR #954, opened 2022-11-24
[solidjs-kit] 22 Implement language filter functionality

hdJerry:
# Implement language filter functionality

## Background

In another ticket, a component was created with the language dropdown button. The purpose of this ticket is to add the functionality and populate the fields for that button. The options that should be available to the user are listed below - when a user selects an option, the repository list on the page should be filtered to only include ones that match the user’s selection.

## Acceptance

- [x] language options should show all (default) and list of languages from users’ repos


vyktoremario (approved):
Looks good.
I am just curious. If I select a language `typescript` and it filters and returns only repos for that language, what will happen if I click on typescript again?

I am expecting it to unselect and go back to default (`All`) which will list all repos. Is this true?

PR #977, opened 2022-11-28
feat(solidjs): Create gists view

SDaian:
## Acceptance Criteria

- [X] Should display Gists views 

## Preview 

image

### DIFF
--- a/solidjs-tailwind/src/pages/Home.jsx
+++ b/solidjs-tailwind/src/pages/Home.jsx
@@ -2,6 +2,7 @@ import { createResource } from 'solid-js';
 import { useAuth } from '../auth';
 import { useOctokit } from '../github';
 import { TopRepos } from '../components/TopRepos';
+import UserGists from '../components/UserGists/UserGists';
 
 const Home = () => {
   useAuth().preventUnauthorised();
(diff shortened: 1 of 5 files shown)


solidjs-tailwind/src/pages/Home.jsx
```diff
@@ -1,15 +1,33 @@
...
 
+  const [data] = createResource(() => {
+    try {
+      return useOctokit()
+        .rest.users.getAuthenticated()
+        .then((response) => response.data);
+    } catch {
+      return Promise.resolve({});
+    }
+  });
```
danecando, line 18: @AllStackDev1 mentioned in the last PR but it looks like we're using the gql api with the fetch util for all the data fetching in this starter kit. We might want to get rid of Ocotokit if thats the case

danecando (commented):
Looks good to me. I left one comment about the Octokit thing. I'm guessing the data fetching will be added to Container in another PR?

solidjs-tailwind/src/pages/Home.jsx
SDaian: yep @danecando that's right, we should get rid of the Octokit, also the PR was rebased within `main` to have the latest commits.

danecando (approved):

AllStackDev1 (approved):
LGTM

PR #979, opened 2022-11-28
feat: [sveltekit pod] missing issues  filters

### LINKED ISSUE #901: Fix Filters For Pull Requests And Issues
# Acceptance Criteria

- The filters for pull requests and issues should allow the user to select filter criteria and have the results update correctly in the UI

ihardz:
Partially resolves #901

Includes: 
* milestone service
* milestone filter
* dropdown adjustments (appearance direction)

### DIFF
--- a/svelte-kit-scss/src/lib/components/IssueSearch/IssueSearchControls/IssueSearchControls.svelte
+++ b/svelte-kit-scss/src/lib/components/IssueSearch/IssueSearchControls/IssueSearchControls.svelte
@@ -9,11 +9,6 @@
   export let stateFilters: NavigationFilterOption[];
   export let milestoneFilters: NavigationFilterOption[];
 
-  const navigateHref = async (href: string): Promise<void> => {
-    await prefetch(href);
-    await goto(href);
-  };
-
   const handleFilterSelect = async (
     option: NavigationFilterOption,
     optional = false
@@ -21,7 +16,7 @@
     if (!optional && option.active) {
       return;
     }
-    await navigateHref(option.href);
+    await goto(option.href);
   };
 </script>
 
--- a/svelte-kit-scss/src/lib/components/shared/Dropdown/DropdownMenu.svelte
+++ b/svelte-kit-scss/src/lib/components/shared/Dropdown/DropdownMenu.svelte
@@ -1,9 +1,10 @@
 <script lang="ts">
-  import { browser } from '$app/environment';
   import { clickOutside } from '$lib/components/shared/actions/clickOutside';
   import { X16 } from 'svelte-octicons';
   import { scale } from 'svelte/transition';
   export let description: string | undefined;
+  type Direction = 'left' | 'right';
+  export let direction: Direction = 'right';
   let isOpen = false;
   const close = (): void => {
     isOpen = false;
@@ -13,8 +14,7 @@
   };
 
   $: descriptionDisplay = description || String();
-  type Direction = 'left' | 'right';
-  export let direction: Direction = 'right';
+
   const reverseDirection = (direction: Direction) => (direction === 'left' ? 'right' : 'left');
 </script>
 
@@ -49,7 +49,7 @@
       font-size: 12px;
       line-height: 1;
       position: absolute;
-      z-index: 9999;
+      z-index: 1;
       top: 100%;
       // left: 0; inline
       margin-top: 8px;


svelte-kit-scss/src/lib/components/IssueSearch/IssueSearchControls/IssueSearchControls.svelte
```diff
@@ -7,27 +7,46 @@
 
   export let sortFilters: NavigationFilterOption[];
   export let stateFilters: NavigationFilterOption[];
+  export let milestoneFilters: NavigationFilterOption[];
 
-  const handleFilterSelect = async (option: NavigationFilterOption): Promise<void> => {
-    if (option.active) {
+  const navigateHref = async (href: string): Promise<void> => {
+    await prefetch(href);
```
BigAB, line 13: I don't think there is any benefit at all to calling  `prefetch(href)` and then calling `goto(href)`.

At this point `goto(href)` would be fine.

svelte-kit-scss/src/lib/helpers/issue-milestone.ts
```diff
@@ -0,0 +1,17 @@
+import {
+  IssueMilestoneState,
+  type GithubIssueMilestone,
+  type IssueMilestone,
+} from '$lib/interfaces';
+
+export const remapIssueMilestone = (milestone: GithubIssueMilestone): IssueMilestone => {
```
BigAB, line 7: What does the `re` part of `remapIssueMilestone` mean?

This function basically validates that `milestone.state` is valid right?

Maybe we just need an `isMilestone` validater function and just pass the API return values through that?

svelte-kit-scss/src/lib/interfaces/issue-milestone.ts
```diff
@@ -0,0 +1,11 @@
+export enum IssueMilestoneState {
```
BigAB, line 1: I think we are over-typing everything here, especially with enums like this, it just causes us to have to import enums all around more.

I wonder if it would just be better to have:
```ts
export interface IssueMilestone {
  id: number;
  number: number;
  title: string;
  state: 'open' | 'closed';
}
```

at the very least we really only ever need one type to represent either `open/closed` across all the other types.

We can merge this as is, and refactor later.

svelte-kit-scss/src/lib/services/issue-milestone-service.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (parameters?.state) {
+      getItemsUrl.searchParams.append(SEARCH_PARAMS.STATE, parameters.state);
+    }
+
+    const response = await this.fetch(getItemsUrl.toString());
+    if (!response.ok) {
+      const body = await response.json();
+      throw error(response.status, body?.message || response.statusText);
+    }
+    const responseBodyJson = (await response.json()) as Promise<GithubIssueMilestone[]>;
```
BigAB, line 43: Generally I think casting the type with `as` is a code smell.  Sometimes it can't be avoided, but it shouldn't be the goto solution.

I would rather see something just validate the api response is of `GithubIssueMilestone` type than casting it.

This does not need to be changed in this PR though, we can refactor later.

svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/[issueSearchType=issue_search_type]/+page.server.ts
```diff
@@ -94,9 +109,25 @@ export const load: PageServerLoad = async ({ fetch, params, url: { searchParams,
...
+    labelParameterDictionaryMilestones,
+    (label) => label,
+    true
+  );
+  console.log(labelParameterDictionaryMilestones);
   return {
     issues,
     sortFilters,
     stateFilters,
+    milestoneFilters,
```
Comment (BigAB) [inline svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/[issueSearchType=issue_search_type]/+page.server.ts:131]: There is a lot going on in this `load()` that looks like it could probably be simplified, but we can address this with refactoring

svelte-kit-scss/src/lib/components/shared/Dropdown/DropdownMenu.svelte
```diff
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { browser } from '$app/environment';
```
BigAB, line 2: Unused?

svelte-kit-scss/src/lib/components/shared/Dropdown/DropdownMenu.svelte
```diff
@@ -12,14 +13,21 @@
   };
 
   $: descriptionDisplay = description || String();
+  type Direction = 'left' | 'right';
+  export let direction: Direction = 'right';
```
BigAB, line 17: As a convention its a good idea to have all the "props" together near the top of the script

svelte-kit-scss/src/lib/components/shared/Dropdown/DropdownMenu.svelte
```diff
@@ -41,8 +49,9 @@
       font-size: 12px;
       line-height: 1;
       position: absolute;
+      z-index: 9999;
```
BigAB, line 52: I would normally advise only ever using `z-index: 1;` or `z-index: -1`, there is little need for any other values

BigAB (changes requested):
Some small changes requested and bigger chanegs noted but that can wait

svelte-kit-scss/src/lib/interfaces/issue-milestone.ts
BigAB: We don't need to match what they do, but also what would be functionally different if we did not use a typescript `enum`.

That schema just describes that the returned value will always be one of `"open"` or `"closed"`, so if in our code we had `state: 'open' | 'closed'` that would be correct would it not?

svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/[issueSearchType=issue_search_type]/+page.server.ts
Comment (BigAB) [reply svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/[issueSearchType=issue_search_type]/+page.server.ts:131]: No, this is fine.

svelte-kit-scss/src/lib/interfaces/issue-milestone.ts
BigAB: It makes sense, but I do think our types are overly verbose.

> What refactoring will be needed this case?

We would just add another string literal `state:  'open' | 'closed' | 'all'` or if we were sharing a `state` type across models, we would just change it.  I don't think this is as big a deal.

But this is still fine as is.

svelte-kit-scss/src/lib/helpers/issue-milestone.ts
BigAB: I will add a suggestion as a future ticket

svelte-kit-scss/src/lib/services/issue-milestone-service.ts
BigAB: I will add a suggestion as a future ticket

BigAB (approved):

Issue #972, opened 2022-11-25
[sveltekit-pod] RepositoryCard.spec.ts: Uncomment the skipped test


kenslachtajr:
svelte-kit-scss/src/lib/components/RepositoryCard/RepositoryCard.spec.ts
In order to expidite merging we commented out a failing test.
- line 38: `it('should should render repo last update', () => {` test is commented out
Please uncomment and fix, or remove test if it is not appropriate

PR #986, opened 2022-11-29
[sveltekit-pod] RepositoryCard.spec.ts: Uncomment the skipped test

iansamz:
Resolves: #972

> File: svelte-kit-scss/src/lib/components/RepositoryCard/RepositoryCard.spec.ts
> In order to expidite merging we commented out a failing test.
> - line 38: `it('should should render repo last update', () => {` test is commented out
> Please uncomment and fix, or remove test if it is not appropriate

### DIFF
--- a/svelte-kit-scss/src/lib/components/RepositoryCard/RepositoryCard.spec.ts
+++ b/svelte-kit-scss/src/lib/components/RepositoryCard/RepositoryCard.spec.ts
@@ -2,6 +2,7 @@ import { beforeEach } from 'vitest';
 import { render, screen } from '@testing-library/svelte';
 import RepositoryCard from './RepositoryCard.svelte';
 import { topRepositoriesFixture } from '$lib/fixtures';
+import { relativeTimeFmt } from '../../helpers';
 
 describe('RepositoryCard', () => {
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


svelte-kit-scss/src/lib/components/RepositoryCard/RepositoryCard.spec.ts
```diff
@@ -35,8 +36,8 @@ describe('RepositoryCard', () => {
     expect(license).toBeTruthy();
   });
 
-  // it('should should render repo last update', () => {
-  //   const lastUpdate = screen.getByText(/weeks ago/);
-  //   expect(lastUpdate).toBeTruthy();
-  // });
+  it('should should render repo last update', () => {
+    const lastUpdate = screen.getByText(relativeTimeFmt('2022-10-25T16:41:51Z'));
```
BigAB, line 40: What was the problem here?

It seems to me like the original text was "correct" in that it was expecting a certain string to appear based on the fixtures, but now we are using an implementation detail in the test, to assert the text shown.

iansamz: The relative time was changing breaking the test. Putting a hard coded string will not work in this  case

BigAB: I see.
So I understand then, the fixtures have a set time, which then displays in the component as relative to "now", which of course changes whenever the test is run.

This is a tricky but common problem.

I think since the `relativeTimeFmt` is a direct dependancy of the `RepositoryCard`, that this use in the test is appropriate, an alternative would've been to mock `relativeTimeFmt` which seems excessive, so that's just a long winded way of saying, this is good.

BigAB (approved):

Issue #969, opened 2022-11-25
[sveltekit-pod] Dropdown: Add aria-attributes to our custom multiselect dropdown


kenslachtajr:
svelte-kit-scss/src/lib/components/shared/Dropdown/item-templates/DropdownItemTemplateCheckbox.svelte
   + related files if required
- I think we need to add some aria listbox attributes to this custom multi select dropdown
- https://www.w3.org/TR/wai-aria-1.1/#listbox
This will demonstrate our commitment to accessibility and show that we are aware of it in our showcases.

PR #987, opened 2022-11-29
[sveltekit-pod] Dropdown: Add aria-attributes to our custom multiselect dropdown

iansamz:
Resolves: #969

> File: svelte-kit-scss/src/lib/components/shared/Dropdown/item-templates/DropdownItemTemplateCheckbox.svelte
>    + related files if required
> - I think we need to add some aria listbox attributes to this custom multi select dropdown
> - https://www.w3.org/TR/wai-aria-1.1/#listbox
> This will demonstrate our commitment to accessibility and show that we are aware of it in our showcases.
>


BigAB (approved):
We are going to need more than this for an accessible dropdown multiselect, but for now this is good, we can improve as a ticket in itself

PR #989, opened 2022-11-29
Solid/fix header not showing on refresh

hdJerry:
Header is not showing are a result of not persisting certain datas. Find a way to persist the token and ensure the data required for the header is fetched accordingly


AllStackDev1 (approved):
LGTM

Issue #990, opened 2022-11-29
Remove Max width on repo header


iansamz:
![Image](https://user-images.githubusercontent.com/15368874/204534803-b0f831f1-12fe-4766-baa7-153d42a91f3b.png)

PR #991, opened 2022-11-29
Remove Max width on repo header

iansamz:
Resolves: #990

Result: 
image

> 
Image

### DIFF
--- a/svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte
+++ b/svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte
@@ -6,9 +6,9 @@
   const { repoInfo } = data;
 </script>
 
-<div class="container">
+<div>
   <RepoHeader repo={repoInfo} />
-  <div class="main">
+  <div class="main container">
     <slot />
   </div>
 </div>
--- a/svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte
+++ b/svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte
@@ -6,17 +6,10 @@
   const { repoInfo } = data;
 </script>
 
-<div>
-  <RepoHeader repo={repoInfo} />
-  <div class="main container">
-    <slot />
-  </div>
+<RepoHeader repo={repoInfo} />
+<div class="container">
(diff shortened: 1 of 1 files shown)


svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte
```diff
@@ -6,9 +6,9 @@
   const { repoInfo } = data;
 </script>
 
-<div class="container">
+<div>
   <RepoHeader repo={repoInfo} />
-  <div class="main">
+  <div class="main container">
```
Comment (BigAB) [inline svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte:11]: If you look at the `<style>` tag, you can see you are maybe having unintended changes.

Now there is no `.main` that is a descendent of `.container` and there is nothing styling or using the `.main` class anymore.

I think this looks fine though so you can probably just remove the `container` class entirely and the whole `<style>` tag, and there is an extra unneeded `<div>` in the template...  it can probably just be

```svelte
<script lang="ts">
  import RepoHeader from '$lib/components/Repo/RepoHeader/RepoHeader.svelte';
  import type { LayoutServerData } from './$types';

  export let data: LayoutServerData;
  const { repoInfo } = data;
</script>

<RepoHeader repo={repoInfo} />
<div class="container">
  <slot />
</div>

```

...and now that I am looking we don't use `repo` or `username` from the server data 🤷‍♂️

BigAB (changes requested):

svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte
```diff
@@ -6,17 +6,10 @@
...
-  <div class="main">
-    <slot />
-  </div>
+  <slot />
 </div>
 
 <style lang="scss">
-  .container {
-    .main {
-      padding: 1.5rem;
```
Comment (BigAB) [inline svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte:19]: Did we need this `padding: 1.5rem` though?

Comment (iansamz) [reply svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte:19]: No it was not needed

Comment (iansamz) [reply svelte-kit-scss/src/routes/(authenticated)/[username]/[repo]/+layout.svelte:19]: No it wasnt needed

BigAB (approved):

PR #1000, opened 2022-11-29
Solidjs home page top repos list

AllStackDev1:
Closes: https://github.com/thisdot/pods/issues/128

image


solidjs-tailwind/src/components/PrivacyBadge/PrivacyBadge.jsx
```diff
@@ -1,12 +1,7 @@
...
+import cn from 'classnames';
 
 const PrivacyBadge = (props) => {
-  const [local] = splitProps(props, ['visibility']);
-  return (
-    <span class="py-0.5 px-2 text-xs rounded-xl text-gray-600 border border-gray-300 font-medium capitalize">
-      {local.visibility.toLowerCase()}
-    </span>
-  );
+  return <span class={cn('py-0.5 px-2 text-xs rounded-xl text-gray-600 border border-gray-300 font-medium', props.class)}>{props.isPrivate ? 'Private' : 'Public'}</span>;
```
hdJerry, line 4: the value here can't be as there are not just private and public.
return to the previous implementation. and probably in the query and API add visibility, you can use the user repo as a guide.

solidjs-tailwind/src/components/PrivacyBadge/PrivacyBadge.stories.jsx
```diff
@@ -3,13 +3,15 @@ import PrivacyBadge from './PrivacyBadge';
 export default {
   title: 'Components/PrivacyBadge',
   argTypes: {
-    visibility: {},
+    isPrivate: true,
```
hdJerry, line 6: no, please. return to the previous implementation.

hdJerry: and by the way, in argTypes, you don't add the value.

solidjs-tailwind/src/components/PrivacyBadge/PrivacyBadge.stories.jsx
```diff
@@ -3,13 +3,15 @@ import PrivacyBadge from './PrivacyBadge';
...
+    className: '',
   },
 };
 
 const Template = (args) => <PrivacyBadge {...args} />;
 
 export const Default = Template.bind({});
 Default.args = {
-  visibility: 'Public',
+  isPrivate: true,
```
hdJerry, line 15: return to previous implementation

solidjs-tailwind/src/components/RepoCard/RepoCard.jsx
```diff
@@ -1,59 +1,48 @@
...
-
   return (
     <div class="px-4 py-4 border-b border-gray-200 first-of-type:border-t flex justify-between flex-wrap md:flex-nowrap gap-x-4">
       <div class="col-span-12 md:col-span-7 text-left">
         <h3 class="mb-4">
-          <Link href={repoNameWithOwnerLink()}>
+          <Link href={`/${local.owner}/${local.name}`}>
             <span class="text-xl text-blue-600 font-semibold hover:underline mr-3">
-              {repoNameWithOwner()}
+              {local.name}
```
hdJerry, line 25: no please, return to the previous implementation. If something isn't clear, please ask why it is done that way.

hdJerry (changes requested):
Worst case scenario is modifying the API or the data going in, not the UI content

solidjs-tailwind/src/components/PrivacyBadge/PrivacyBadge.jsx
vyktoremario: The API returns `isPrivate`. So you can do:

`{local.isPrivate ? 'Private' : 'Public'}`

hdJerry (approved):
Thanks

PR #1037, opened 2022-12-07
Fix Vitest $app modules support

iansamz:
Resolves: #1036


svelte-kit-scss/vite.config.ts
```diff
@@ -1,10 +1,38 @@
...
+import { configDefaults, type UserConfig as VitestConfig } from 'vitest/config';
+import path from 'path';
 
-const config: UserConfig = {
+const config: UserConfig & { test: VitestConfig['test'] } = {
   plugins: [sveltekit()],
   define: {
-    global: '({})',
+    // Eliminate in-source test code
+    'import.meta.vitest': 'undefined',
```
BigAB, line 10: Oh, why do this? I am curious.

svelte-kit-scss/setupTests.ts
```diff
@@ -0,0 +1,76 @@
...
+vi.mock('$app/environment', (): typeof environment => ({
+  browser: false,
+  dev: true,
+  building: false,
+  version: 'any',
+}));
+
+// Mock SvelteKit runtime module $app/navigation
+vi.mock('$app/navigation', (): typeof navigation => ({
+  afterNavigate: () => {},
```
BigAB, line 23: I think if you just called `vi.mock('$app/navigation')` you would get the same result

iansamz: Let me give it a try

iansamz: I believe this is necessary as some navigation types need to resolve async and others do not

svelte-kit-scss/vite.config.ts
BigAB: Ohh, I see.

I actually think in-source testing seems like a good idea

BigAB (approved):
LGTM :shipit:

PR #1053, opened 2022-12-13
fix: Netlify backend axios

markshenouda:
Replacing Axios with Node fetch because it doesn't work with Netlify (We don't know why) as you see in the next screenshot we logged the result of the data in axios request using `console.log('Result:', result.data);` and this was the result

image

it looks like encoded, we tried to set the `responseEncoding` to `utf-8` but no result

This is the result of logging the whole result

image

this is the same request but with `node-fetch`

image

the last one looks right, so I removed the axios package and migrated to node-fetch


Zelig880 (approved):
Code seem reasonable but i have not tested this locally!

i am going to merge this to hopefully unblock people

PR #1055, opened 2022-12-14
Implementing brower storage for caching

AllStackDev1:
This PR was inspired by the blog post I wrote here: https://docs.google.com/document/d/1njNHhv7PGnHN1o0uC_mc3sURZlH4NbP4vD7HhUza5Nw/edit?usp=sharing


danecando (commented):
It doesn't look like there's any kind of invalidation for the cache. Not sure we would want data cached indefinitely (until the users session storage clears)

PR #1066, opened 2022-12-14
fix(svelte-kit-scss): replace api url in example.env and remove alias…

BigAB:
The PR updates the `env.example` file with the new API endpoint and removes an unnecessary alias that SvelteKit was warning about:
 
Screenshot 2022-12-14 at 1 07 41 PM


iansamz (approved):
LGTM

PR #1157, opened 2022-12-29
[svelte-kit-scss] Add empty state on search on profile

### LINKED ISSUE #1124: [svelte-kit-scss] Add empty state on search on profile

ihardz:
Resolves #1124

### DIFF
--- a/svelte-kit-scss/src/routes/(authenticated)/[username]/+page.server.ts
+++ b/svelte-kit-scss/src/routes/(authenticated)/[username]/+page.server.ts
@@ -2,7 +2,6 @@ import type { PageServerLoad } from './$types';
 import {
   buildRepositoryCardViewModel,
   buildRepositoryPageNavigationFilterOptions,
-  buildURLSearchParamsForParameters,
   DEFAULT_REPOSITORY_SEARCH_QUERY_PARAMETERS_REQUIRED,
   extractRepositoryPageSearchQueryParameters,
   isRepositorySearchQueryParametersEqual,


svelte-kit-scss/src/lib/components/RepositoryList/AllRepositoriesList/AllRepositoriesList.svelte
```diff
@@ -1,15 +1,28 @@
...
-  <!-- todo: add pagination here (slot="bottom")-->
+  {#if hasItemsToShow}
+    <RepositoryList {items} />
+    <!-- todo: add pagination here (slot="bottom")-->
+  {/if}
 </BoxLayout>
+
+{#if !hasItemsToShow}
+  {#if hasSelectedFilters}
+    <BlankSlate heading="The profile doesn't have any repositories that match." />
```
mbicknese, line 24: Is the `BlankSlate` supposed to live outside of the `BoxLayout`?

svelte-kit-scss/src/lib/components/shared/layouts/BoxLayout.svelte
```diff
@@ -35,6 +40,9 @@
       border: $border;
       border-radius: 6px;
       background-color: variables.$white;
+      &.empty {
```
mbicknese, line 43: This looks like a fix to a problem we shouldn't have. As your solution above demonstrates, we typically know when the layout is empty. In this case, we can prevent rendering the whole layout altogether.

Just my take glancing at this code. I might be missing things.

mbicknese (changes requested):
I have two questions.
There's also a lint warning which is easily fixed.

mbicknese (approved):
Thanks for clarifying the changes. LGTM 🚢

PR #1158, opened 2022-12-30
feat(qwik-pod): filter function on issues page

Megio:
Added milestone and label filter on issue page

Fixes: #1149

### DIFF
--- a/qwik-graphql-tailwind/src/components/issue-pr-card/card.tsx
+++ b/qwik-graphql-tailwind/src/components/issue-pr-card/card.tsx
@@ -80,15 +80,16 @@ export const IssuePrCard = component$(({ data, type }: IssuePrCardProps) => {
           <a class="align-middle no-underline markdown-title font-semibold" target="_blank" href={data.url}>
             {data.title}
           </a>
-          {data.labels?.map((label: Label) => (
-            <span
-              key={label.color}
-              class={cn('mt-2 ml-2 py-1 px-2 rounded-full text-sm', `bg-[#${label.color}]`)}
-              style={{ backgroundColor: `#${label.color}` }}
-            >
-              {label.name}
-            </span>
-          ))}
+          {data.labels &&
+            data.labels.map((label: Label) => (
+              <span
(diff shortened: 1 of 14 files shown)


qwik-graphql-tailwind/src/components/issue-tab-view/issue-tab-view.tsx
```diff
@@ -45,22 +51,35 @@ export const IssueTabView = component$(({ activeTab, owner, name }: IssuesProps)
     activeTab: activeTab,
   });
   const dropdownStore = useStore<DropdownStoresProps>({
-    selectedLabel: labelOptions[0].value,
+    selectedLabel: undefined,
```
iamjerryhogan, line 54: why undefined not null?

qwik-graphql-tailwind/src/components/issue-tab-view/issue-tab-view.tsx
```diff
@@ -45,22 +51,35 @@ export const IssueTabView = component$(({ activeTab, owner, name }: IssuesProps)
...
     selectedSort: sortOptions[0].value,
-    selectedMilestones: milestonesOptions[0].value,
+    selectedMilestones: undefined,
```
iamjerryhogan, line 56: why undefined and not null?

Megio: in order to have the first "no milestone selected" value checked when we open the filter dropdown. We check the value of the option with the dropdown store value and the initial value of the store has no "selectedMilestone" so is computed as undefined, not null. (Don't know if I was clear, sorry)

qwik-graphql-tailwind/src/components/issue-tab-view/issue-tab-view.tsx
Megio: to be consistent with the milestone value, that needs to be undefined

SDaian (approved):
LGTM 🤘🏼

PR #1160, opened 2023-01-03
fix: file nav explorer for file viewer

iansamz:


iansamz:
Currently next button is not working

PR #1184, opened 2023-01-06
[sveltekit-pod] User profile unit tests

iansamz:
Tests include:
1. OrganizationProfile
2. UserProfile
3. OrganizationMembers
4. OrganizationDetails


svelte-kit-scss/src/lib/components/Profile/OrganizationProfile/OrganizationProfile.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    ['bio', mockOrgProfile.bio],
+    ['followers', mockOrgProfile.followers],
+    ['email', mockOrgProfile.email],
+    ['blog', mockOrgProfile.blog],
+    ['twitterUsername', mockOrgProfile.twitterUsername],
+    ['location', mockOrgProfile.location],
+  ])('should render: %s', (testId, expectedValue) => {
+    const element = screen.getByTestId(testId);
+    const expectedText = String(expectedValue);
+    expect(element.innerHTML).toEqual(expectedText);
```
AllanJeremy, line 27: Could we instead use `textContent` to test for this? `innerHTML` would fail this if for whatever reason some elements wrapped the actual text eg. `<strong></strong>`. Also, `innerHTML` may treat certain values as special characters eg. double space could be `&nbsp;`

svelte-kit-scss/src/lib/components/Profile/OrganizationProfile/OrganizationProfile.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+const mockOrgProfile = remapPublicProfileInformation(MOCK_GITHUB_ORGANIZATION_PROFILE_INFORMATION);
+
+describe('OrganizationProfile', () => {
+  beforeEach(() => {
+    render(OrganizationProfile, {
+      profile: mockOrgProfile,
+    });
+  });
+
+  it.each([
```
AllanJeremy, line 16: I think it might help to split these into individual tests since we may want to have different validations on the different fields. Eg. it displays a correctly formatted email or the number of followers is actually a number.

svelte-kit-scss/src/lib/components/Profile/UserProfile/UserProfile.spec.ts
```diff
@@ -20,15 +20,20 @@ describe('UserProfile', () => {
...
+    ['followers', mockUserProfile.followers],
+    ['following', mockUserProfile.following],
+    ['email', mockUserProfile.email],
+    ['blog', mockUserProfile.blog],
+    ['twitterUsername', mockUserProfile.twitterUsername],
+    ['location', mockUserProfile.location],
   ])('should render: %s', (testId, expectedValue) => {
     const element = screen.getByTestId(testId);
     const expectedText = String(expectedValue);
     expect(element.innerHTML).toEqual(expectedText);
```
AllanJeremy, line 32: Same suggestion, prefer `textContent` over `innerHTML`

svelte-kit-scss/src/lib/components/Profile/UserProfile/UserProfile.spec.ts
```diff
@@ -20,15 +20,20 @@ describe('UserProfile', () => {
...
-    // ['location', mockUserProfile.location],
+    ['followers', mockUserProfile.followers],
+    ['following', mockUserProfile.following],
+    ['email', mockUserProfile.email],
+    ['blog', mockUserProfile.blog],
+    ['twitterUsername', mockUserProfile.twitterUsername],
+    ['location', mockUserProfile.location],
   ])('should render: %s', (testId, expectedValue) => {
     const element = screen.getByTestId(testId);
     const expectedText = String(expectedValue);
```
AllanJeremy, line 31: Why do we manually "cast" this to string? Is there behaviour that would have this not being a string?

svelte-kit-scss/src/lib/components/Profile/UserProfile/UserProfile.spec.ts
```diff
@@ -20,15 +20,20 @@ describe('UserProfile', () => {
     ['name', mockUserProfile.name],
     ['login', mockUserProfile.login],
     ['bio', mockUserProfile.bio],
-    // ['followers', mockUserProfile.followers],
-    // ['following', mockUserProfile.following],
-    // ['email', mockUserProfile.email],
-    // ['blog', mockUserProfile.blog],
-    // ['twitterUsername', mockUserProfile.twitterUsername],
-    // ['location', mockUserProfile.location],
+    ['followers', mockUserProfile.followers],
```
AllanJeremy, line 23: Once again, imo, these should be separate tests (if the validation of the different fields would matter)

svelte-kit-scss/src/lib/components/Profile/common/SocialLink.svelte
```diff
@@ -5,6 +5,7 @@
   export let label: string | null | undefined = undefined;
   export let iconComponent: ComponentType;
   export let mobileHide = false;
+  export let testId: string | null | undefined = undefined;
```
AllanJeremy, line 8: `export let testId?: string | null` instead? Noticed that both the current syntax and this are used across the codebase, is it possible to pick a single approach? The suggestion here feels a bit more succinct

svelte-kit-scss/src/routes/(authenticated)/[username]/+page.svelte
```diff
@@ -23,6 +23,7 @@
     <LayoutPageHeader gray>
       <LayoutPageContentRow>
         <div class="org-profile">
+          {console.log(profile)}
```
Comment (AllanJeremy) [inline svelte-kit-scss/src/routes/(authenticated)/[username]/+page.svelte:26]: Unnecessary `console.log`

AllanJeremy (dismissed):
Left a couple comments, generally looks good :)

svelte-kit-scss/src/lib/components/Profile/OrganizationProfile/OrganizationDetails/OrganizationMembers.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+describe('OrganizationMembers', () => {
+  it('should render organization members', () => {
+    render(OrganizationMembers, {
+      members: mockOrgMembers,
+    });
+
+    const links = screen.getAllByRole('link');
+    const images = screen.getAllByRole('img');
+
+    mockOrgMembers.map((member, i) => {
```
mbicknese, line 17: Although it typically works to create assertions dynamically in tests, it is not a good practice because the system doesn't know what to expect when starting the test. Instead, we could combine `beforeAll` and `it.each` to render the `OrganizationMembers` and have a "separate" test for each member. With each test having the two expects.

A small side note, if you don't care about the results of a `map`, you can use `forEach` instead.

svelte-kit-scss/src/lib/components/Profile/OrganizationProfile/OrganizationProfile.spec.ts
```diff
@@ -0,0 +1,62 @@
+import { beforeEach, it, expect } from 'vitest';
+import { render, screen } from '@testing-library/svelte';
+import OrganizationProfile from './OrganizationProfile.svelte';
+import { MOCK_GITHUB_ORGANIZATION_PROFILE_INFORMATION } from '$lib/helpers/mocks/users';
+import { remapPublicProfileInformation } from '$lib/helpers';
+
+const mockOrgProfile = remapPublicProfileInformation(MOCK_GITHUB_ORGANIZATION_PROFILE_INFORMATION);
+
+describe('OrganizationProfile', () => {
+  beforeEach(() => {
```
mbicknese, line 10: Seeing we're not modifying or interacting with the component, we can use `beforeAll` here to speed up the tests.

svelte-kit-scss/src/lib/components/Profile/OrganizationProfile/OrganizationProfile.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    expect(element.textContent).toEqual(String(twitterUsername));
+  });
+
+  it('should render location of org correctly', () => {
+    const element = screen.getByTestId('location');
+    expect(element.textContent).toEqual(String(location));
+  });
+
+  it('should render email of org correctly', () => {
+    if (email) {
```
mbicknese, line 45: I suggest we just make sure there's an email to test for so we don't need this if-statement.

svelte-kit-scss/src/lib/components/Profile/UserProfile/UserProfile.spec.ts
```diff
@@ -16,19 +16,55 @@ describe('UserProfile', () => {
     });
   });
 
-  it.each([
```
mbicknese, line 19: I'm surprised to see you removed this in favour of the non-dynamic setup. What made you move in this direction?

mbicknese (changes requested):

svelte-kit-scss/src/lib/components/Profile/OrganizationProfile/OrganizationDetails/OrganizationMembers.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+      members,
+    });
+  }
+
+  beforeEach(() => {
+    organizationMembersTree();
+  });
+
+  let currentTestIndex = 0;
+  it.each(members)(`should render member ${currentTestIndex}`, ({ href, avatarUrl }) => {
```
mbicknese, line 20:
```suggestion
  it.each(members)(`should render member %#`, ({ href, avatarUrl }) => {
```

You can use `%#` to inject the index into the test name, see https://vitest.dev/api/#test-each

svelte-kit-scss/src/lib/components/Profile/OrganizationProfile/OrganizationDetails/OrganizationMembers.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+describe('OrganizationMembers', () => {
+  function organizationMembersTree() {
+    return render(OrganizationMembers, {
+      members,
+    });
+  }
+
+  beforeEach(() => {
+    organizationMembersTree();
+  });
```
mbicknese, line 17:
```suggestion
  beforeAll(organizationMembersTree);
```

We can use `beforeAll` here to prevent unnecessary rerendering.

mbicknese (changes requested):
Give me a second while I check this branch locally and I'll get back to you with (hopefully) some insightful suggestions.

mbicknese:
Getting links the proper testing-library way, without actual text content, is not as easy as it seemed. Lesson learned.

For the `OrganizationMembers` test, what do you think of the following setup

```javascript
  it.each(members)(`should render member %#`, ({ href, avatarUrl , login}) => {
    const container = render(OrganizationMembers, {
      members
    });
    const image = container.getByAltText(login);
    expect(image).toHaveProperty("src", avatarUrl);
    expect(image.closest('a')?.href).toBe(window.location.origin + href);
  });
```

I feel this is closer to how test-library is supposed to be used (targeting what users can see), and prevent workarounds to keep track of the index.

svelte-kit-scss/src/lib/components/Profile/OrganizationProfile/OrganizationDetails/OrganizationMembers.spec.ts
iansamz: I couldnt get beforeAll working. I get an error on the second item in the array.

mbicknese: My bad, typically beforeAll would work when you want to set something up that is consistent throughout all tests. But the "screen" resets after each test, so we need to rerender.

mbicknese (approved):

Issue #1181, opened 2023-01-05
[solidjs-tailwind] Fix ordering of files and folders when viewing a repo


kenslachtajr:
# Description
The files and folders are currently not listed in the correct order when viewing repo contents. Directories should be first, and individual files should be listed next.

![Image](https://user-images.githubusercontent.com/1828602/210889173-a7d9ca62-b64d-41a0-8501-cc573b5063c9.png)

PR #1186, opened 2023-01-06
feat(solidjs-pod): fix order on File Explorer

SDaian:
## Description

- [x] The files and folders are currently not listed in the correct order when viewing repo contents. Directories should be first, and individual files should be listed next.

Closes: #1181 

## Screenshot

image


hdJerry (approved):
Thanks captain!

PR #1206, opened 2023-01-10
nested folder nav fixed

hdJerry:
# Description
- Issue happens when you navigate into nested folder directories (see screenshot below). Unable to see contents of the clicked folder and back arrow navigation does not function properly. Manually refreshing the page does fix the issue. When clicking a file to view its contents, the URL updates, but the page doesn't update. This is why refreshing the page works because it then forces the current URL in the address bar to load.

Image

### DIFF
--- a/qwik-graphql-tailwind/src/components/file-explorer/index.tsx
+++ b/qwik-graphql-tailwind/src/components/file-explorer/index.tsx
@@ -40,12 +40,17 @@ export const FileExplorer = component$(({ tree }: { tree: any[] }) => {
                   <DocumentIcon className="w-5 h-5 text-gray-500" />
                 )}
               </div>
-              <Link href={`${basePath}/${item.type}/${globalStore.branch || pathBranch}/${item.path}`}>
+              <a href={`${basePath}/${item.type}/${globalStore.branch || pathBranch}/${item.path}`}>
                 <span class="hover:text-blue-600 hover:underline">{item.name}</span>
-              </Link>
+              </a>
             </div>
           </div>
         ))}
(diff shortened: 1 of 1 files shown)


qwik-graphql-tailwind/src/components/file-explorer/index.tsx
```diff
@@ -40,12 +40,17 @@ export const FileExplorer = component$(({ tree }: { tree: any[] }) => {
                   <DocumentIcon className="w-5 h-5 text-gray-500" />
                 )}
               </div>
-              <Link href={`${basePath}/${item.type}/${globalStore.branch || pathBranch}/${item.path}`}>
+              <a href={`${basePath}/${item.type}/${globalStore.branch || pathBranch}/${item.path}`}>
```
markshenouda, line 43: This gonna work only on localhost not in Netlify because we are using the static adaptor, when you're free I can test it with you! but it's fine to leave this bug without a fix if we didn't find a solution, Dustin told us to add a disclaimer in the readme file until the Qwik team fixes the Netlify adaptor

AllStackDev1: This is true, on dev the <Link> has some issues, it doesn't work for some deeply nested routes, but on production and in preview mode the Link works. so I would suggest we make this known so that we don't future create tickets like this, which will result in back and forth.

hdJerry: If that is the case, there is a better work around, we can use qwik useNavigation. be right back!!

PR #1211, opened 2023-01-10
[svelte-kit-scss] HTML in README isn't parsing as code block

### LINKED ISSUE #1198: [svelte-kit-scss] HTML in README isn't parsing as code block
Image

kodejuice:
Closes #1198

### DIFF
--- a/svelte-kit-scss/src/lib/helpers/repository-contents.ts
+++ b/svelte-kit-scss/src/lib/helpers/repository-contents.ts
@@ -67,9 +67,10 @@ export const buildMarkdownPreviewHtml = (file: GithubFileContentsItem): string =
   }
   const decoded = Buffer.from(content, bufferEncoding).toString('utf8');
   return sanitizeHtml(md.render(decoded), {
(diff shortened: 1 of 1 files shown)


mbicknese (changes requested):
This is a good start as it is now no longer showing the raw HTML.

However, the links are getting dropped. I just see the rendered images. Maybe `MarkdownIt` has some safety features which drop a-tags. We might also ask @kenslachtajr whether it's important enough to fix the links.

kodejuice:
> This is a good start as it is now no longer showing the raw HTML.
> 
> However, the links are getting dropped. I just see the rendered images. Maybe `MarkdownIt` has some safety features which drop a-tags. We might also ask @kenslachtajr whether it's important enough to fix the links.

Yes, it needs to be manually added, done that now

iansamz:
This looks OK but I wanted to switch to `mdsvex` as it comes with more benefits: checkout #1240

iansamz (approved):

kodejuice:
> This looks OK but I wanted to switch to `mdsvex` as it comes with more benefits: checkout #1240

yes i saw that, i think i can close this PR, since `mdsvex` solves both problems

Issue #997, opened 2022-11-29
[Sveltekit-pod] Implement unit test for all components


Zelig880:
As an user I would like all the components in the components folder to have unit tests. 

NOTE: This ticket is going to be worked on from multiple people and will have multiple PRs

AC: (add you name next to the component that you are working on)
- [ ] Example component -  Simone
- [x] Error Page - Ian
- [x] FileExplorer - Allan
- [x] FileViewer - Ian
- [x] Gists - Ian
- [x] IssueSearch - Allan
- [x] NavBar - Ian
- [x] PRHeader - Ian
- [x] PRIssueCard - Ian
- [x] PRsList - Ian
- [x] Profile - Ian
- [x] RepoHeader - Ian
- [ ] RepoList - Ian
- [x] RepositoryCard - Ian
- [x] TopRepositories - Ian
- [ ] auth - Ian
- [ ] shared

PR #1220, opened 2023-01-11
[sveltekit-pod] Unit Test Auth

iansamz:
Partly resolves: #997

### DIFF
--- a/svelte-kit-scss/src/lib/components/auth/Auth.spec.ts
+++ b/svelte-kit-scss/src/lib/components/auth/Auth.spec.ts
@@ -1,0 +1,15 @@
+import { render, screen } from '@testing-library/svelte';
+import userEvent from '@testing-library/user-event';
+import Auth from './Auth.svelte';
+
+describe('Auth Component', () => {
+  it('should render sign out', async () => {
+    render(Auth);
+
+    const signInBtn = screen.getByRole('button');
+    expect(signInBtn.textContent).toEqual('Sign in with GitHub');
+    await userEvent.click(signInBtn);
+
+    expect(document.location.href).toEqual('http://localhost:3000/');
+  });
+});
--- a/svelte-kit-scss/src/lib/components/auth/Auth.spec.ts
+++ b/svelte-kit-scss/src/lib/components/auth/Auth.spec.ts
@@ -3,7 +3,7 @@ import userEvent from '@testing-library/user-event';
 import Auth from './Auth.svelte';
 
 describe('Auth Component', () => {
-  it('should render sign out', async () => {
+  it('should render sign in', async () => {
     render(Auth);
 
     const signInBtn = screen.getByRole('button');


svelte-kit-scss/src/lib/components/auth/Auth.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { render, screen } from '@testing-library/svelte';
+import userEvent from '@testing-library/user-event';
+import Auth from './Auth.svelte';
+
+describe('Auth Component', () => {
```
AllanJeremy, line 5: This feels like it is missing additional tests
- Does a user actually get signed in? We can mock this

svelte-kit-scss/src/lib/components/auth/Auth.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+  it('should render sign in', async () => {
+    render(Auth);
+
+    const signInBtn = screen.getByRole('button');
+    expect(signInBtn.textContent).toEqual('Sign in with GitHub');
```
AllanJeremy, line 10: This test is too heavily tied to what the text of the button is, which is an implementation detail. In this case, if we renamed that to "Login with Github" then the test would fail, even though the actual functionality works.

## Suggestion
Rather than testing for equality of the text button's text content, simply check if we have a sign-in button by for example using `data-testid="button-signin"` which would remain constant even if the actual content of the sign-in button changed.

svelte-kit-scss/src/lib/components/auth/Auth.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+    await userEvent.click(signInBtn);
+
+    expect(document.location.href).toEqual('http://localhost:3000/');
```
AllanJeremy, line 13: This test would not work in a CI environment or production/staging environments if you ever needed tests there for whatever reason since the path would potentially not be `localhost`. Instead you could check for the relative path that is returned.

## Suggestion
Use `window.pathname` to check if it returns us to our root application path (`/`).

svelte-kit-scss/src/lib/components/auth/Auth.spec.ts
AllanJeremy: Additionally, @iansamz , what happens if the signin fails? Is it handled gracefully? We should have tests for this to assert that that is indeed the behaviour.

AllanJeremy (changes requested):
Left a bunch of suggestions :)

svelte-kit-scss/src/lib/components/auth/Auth.spec.ts
flakolefluk: using testid is discouraged, and should be used when it's not possible to get a node in a different way. the text in the button is what matters to the user. getByRole is the right choice in this case. you can use the options to make the query precise, and also be more flexible on the text context.
queries are assertions too, it will throw if it finds more than one node or none.
replacing

```js
const signInBtn = screen.getByRole('button');
expect(signInBtn.textContent).toEqual('Sign in with GitHub'); 
```

with

```js
const signInBtn = screen.getByRole('button', { name: /sign in with github/i });
```

should be enough to assert that the button exists, and has more flexibility with the text in it. a change in casing doesn't really matter to the user. It's also resilient to adding more buttons.
Hope it helps.

PR #1245, opened 2023-01-13
[svelte-kit-scss] P1: Fix 500 when clicking into files issue

### LINKED ISSUE #1092: [svelte-kit-scss] P1: Fix 500 when clicking into files issue

ihardz:
resolves #1092

NOTE: was reproduced only on MAC

### DIFF
--- a/svelte-kit-scss/src/lib/helpers/file.ts
+++ b/svelte-kit-scss/src/lib/helpers/file.ts
@@ -1,5 +1,6 @@
 import type { FileContents, GithubFileContentsItem } from '$lib/interfaces';
 import type { Language } from '$lib/components/FileViewer/types/language';
+import { decodeFileContent } from './repository-contents';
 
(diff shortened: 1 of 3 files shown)


mbicknese (approved):
Shipt it ⛵

kodejuice (approved):

svelte-kit-scss/src/lib/helpers/file.ts
```diff
@@ -86,8 +87,7 @@ export const remapFileContents = (
 ): FileContents | undefined => {
   return responseData
     ? {
-        // content: Buffer.from(responseData.content, 'base64').toString('binary'),
```
iansamz, line 89: Would like to request if we can use Buffer.from had changed it with Adam trying to fix adapter but it did not solve the issue. You can revert to BUffer

iansamz (changes requested):

PR #1251, opened 2023-01-16
fix: lint warnings

iansamz:
Resolves: #1250


mbicknese (changes requested):
This solution is only hiding the issue from eslint. The actual issue of it not being a valid href persists.

We could replace the `a` tags with `button` if we want to have the impression of the element being interactive without it actually doing anything. Or remove the interactive part completely, as there's no interaction.

PR #1261, opened 2023-01-16
fix: build errors

iansamz:
Resolves: #1260 & #1251


mbicknese (changes requested):
Same as with #1251 for the a11y anchor tags.

And could you add a newline to the end of `netlify.toml`?

mbicknese (approved):
Ship it ⛵

PR #1279, opened 2023-01-17
qwik/fix-page-reload

hdJerry:


Megio (commented):
The changes you made looks good. But when we clear the filters in issues and pr pages, the reload is faster but all page still disappear and reload. I think that probably is due to this line on every resetFilter function we have (issue-tab-view and repo-pulls) and in toggleTab (pull-request-issue-tab component):

`window.location.href = ${pathname}?tab=${value}`

Using window.location all page is reloaded. Why do we need this? I think we could remove this line and everything will still work fine.

SDaian (approved):
LGTM

Megio (approved):
LGTM 🚀

PR #1291, opened 2023-01-18
Solidjs: filter option works as expected

hdJerry:
# Description
- filters should be able to be reset, should be able to click on a selected filter and deselect it after it's been selected, or have the option to Select all.
- when clicking the "X" none of the filters or filtered results are cleared. It doesn't appear to be doing anything currently.

Image

## Result

How it works is when you click on a filter option twice it resets to the default, for example; for sort option the default is newest while for label option the default on nothing..


Megio (dismissed):
LGTM 🚀

Megio (approved):
🙌🏻

PR #1302, opened 2023-01-19
[Next] Fix: Add Netlify badge to sign in screen

lindakatcodes:
Initially, we were using the default sign in screen that Next Auth provided for us, and with that we didn't have a way of adding the Netlify badge on that screen. So this PR handles creating a custom sign in screen that Next Auth can use, that we can add the badge onto.

Since I needed to create a new page, I also had to make a few adjustments so that things redirected as a user would expect, and also hid the nav bar on the sign in page so it's not confusing. Becuase the nav bar wouldn't hide completely, I made the background of the sign in page the same dark blue color to match it. 

I've tested everything locally and it all appears to be working - users can sign in and navigate as expected.

Side note: the es-lint ignore rule is because Next wants us to use their `Image` component instead of the HTML `img` element, but doing so requires us to add a width & height or display property to the badge, which didn't seem right to me since this is directly what Netlify asks us to use.

Screenshots of the new page (the little red icon in the bottom left is the React Query devtools icon, and doesn't show on production):
Screen Shot 2023-01-19 at 3 44 54 PM

Screen Shot 2023-01-19 at 3 48 11 PM

### DIFF
--- a/next-react-query-tailwind/src/pages/auth/signin.tsx
+++ b/next-react-query-tailwind/src/pages/auth/signin.tsx
@@ -1,0 +1,41 @@
+/* eslint-disable @next/next/no-img-element */
+import { CtxOrReq, getProviders, signIn } from 'next-auth/client';
+import { InferGetServerSidePropsType } from 'next';
+
+export default function SignIn({
+  providers,
+}: InferGetServerSidePropsType<typeof getServerSideProps>) {
+  return (
+    <>
+      {providers
+        ? Object.values(providers).map((provider) => (
+            <div key={provider.name}>
+              <button onClick={() => signIn(provider.id, { callbackUrl: '/' })}>
+                Sign in with {provider.name}
+              </button>
+            </div>
+          ))
+        : ''}
+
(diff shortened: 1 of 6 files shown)


next-react-query-tailwind/src/pages/auth/signin.tsx
```diff
@@ -0,0 +1,50 @@
...
+              >
+                <button
+                  className="self-center w-3/5 lg:w-1/5 py-5 px-2 text-white border-solid border-white border-2   rounded-sm hover:text-gray-900 hover:bg-white transition-colors"
+                  onClick={() => signIn(provider.id, { callbackUrl: '/' })}
+                >
+                  Sign in with {provider.name}
+                </button>
+              </div>
+            ))
+          : ''}
```
danecando, line 25: just a nit pick I would use null instead of an empty string but I'm not sure it makes a difference in practice

danecando (dismissed):
LGTM working locally

danecando (approved):

PR #1309, opened 2023-01-23
Solidjs/fix storybook error

hdJerry:
# Description
- When running `yarn storybook`

Image

Due to the fact we a using one API url and also some of the components are wrapped in a context that makes an API call, its not possible to interfere as wanted. But no more errors. If we wanna ensure the mocked data shows in storybook for some cases, we might need to restructure the code.

Screenshot 2023-01-23 at 11 23 06


Megio (commented):
File Explorer Story remains in loading status.
It seems that something is missing in the mocked response probably.

Screenshot 2023-01-23 at 15 54 10

hdJerry:
> File Explorer Story remains in loading status. It seems that something is missing in the mocked response probably.
> 
> Screenshot 2023-01-23 at 15 54 10

Not really, its because of the API URL and the RepoProvider.

Megio (approved):
LGTM

PR #1311, opened 2023-01-23
added loader skeleton component

hdJerry:
At the moment there is a Loading... text when every data is being fetched which make the user experience bad. Please add a loader skeleton in stead of the Loading... text


Megio (approved):
LGTM 🚀

PR #1315, opened 2023-01-23
fixed repo card and test

hdJerry:
Repo card in the main page should only show the repo name not with the username or org name.

Shouldn't be like this

Image

but like this

Image


Megio (approved):
LGTM

Issue #1313, opened 2023-01-23
[solidjs-tailwind] Make Code the active tab when navigating


kenslachtajr:
# Description
- When navigating through nested folders or files, the active tab should still be Code, but for some reasons, it no longer shows the "active" indicator.

![Image](https://user-images.githubusercontent.com/1828602/214056068-778f2259-e7b3-443d-adb2-de27f55bc712.jpg)
![Image](https://user-images.githubusercontent.com/1828602/214056067-e5ea0404-94c9-4055-b0b0-f3ff94fd7267.jpg)

PR #1316, opened 2023-01-23
solid-fix: header tab in repo page 

Megio:
Fixed header tab highlight in repo page and fixed no result behaviour in issues and prs page (now the header is visible also when we have no results and filter could be deselected)

Fixes: #1313


hdJerry (approved):
Works

PR #1317, opened 2023-01-23
fix: token not being set during initial login

### LINKED ISSUE #1292: [cra-rxjs] gets stuck in loading view when first logging in
When first logging into the showcase, the site gets stuck in the loading view. If you navigate to the profile page everything loads, and then any other page (even the home page) after that works as expected. So something on that first initial load is handing. 

Screenshot here shows the calls that are showing as failing in the console:

Image

jdwilkin4:
This PR fixes the issue where users are in a constant loading state when first logging in.

closes #1292 

## Video of fix

https://user-images.githubusercontent.com/67210629/214130274-dda1f082-806a-4d94-b89a-c138b6a00c74.mov


lindakatcodes (approved):
Working locally and looks like it makes sense. 👍🏼

PR #1330, opened 2023-01-25
html file shows now

hdJerry:
# Description
- When clicking an HTML file to view its contents, a 404 error is currently being returned

Image

## Result

Screenshot 2023-01-25 at 10 34 57


SDaian (approved):
LGTM 🚀

Issue #1320, opened 2023-01-24
[solidjs-tailwind] Reload content and not entire page


kenslachtajr:
# Description 
- for issues and profile pages the page shows blank for a moment before loading when navigating to a new route

PR #1331, opened 2023-01-25
solid-fix: manage loading status on every pages

Megio:
Fixed loading status on every page.

Now when we navigate through repo folders and file, only the page's content is loaded and the header remains visible.

Also, in Profile page, when we set and clear filters only the list of repos is loaded and not the whole page.

Fixes: #1320


SDaian (approved):
LGTM

PR #1335, opened 2023-01-26
changed the review port to 3000 in vite config

hdJerry:
# Description
- there is an issue when running `yarn serve`. It goes to port 4173 but then when you authenticate or try to sign in it goes to port 3000. The redirect path is probably wrong and needs to be updated

## Solution
- configured preview port in vite config to port 3000
Then run this command. You have to run yarn build before you run yarn serve
```bash 
yarn build && yarn serve
```


Megio (approved):
LGTM 🚀
For me, it worked also without running yarn build before yarn serve

hdJerry:
> LGTM 🚀 For me, it worked also without running yarn build before yarn serve

You probably have a `dist` folder.

Issue #1341, opened 2023-01-31
[solidjs-tailwind] Fix label filters not static once value is selected issue in showcase


kenslachtajr:
# Description 
- The issue with “label filters not static once the value is selected”. We need to figure out how to fix this using results caching. We need to find this knowledge and get it fixed because this is a common problem we should be able to handle and has been reoccurring on other showcases.

![Image](https://user-images.githubusercontent.com/1828602/215667513-14ef1dcc-f2eb-499b-9f62-bf65b3dd63e9.png)

![Image](https://user-images.githubusercontent.com/1828602/215667554-9ebdcd6c-4d6e-40e3-9e15-cc06902e3901.png)

PR #1348, opened 2023-01-31
Solidjs-fix: label filter

Megio:
Now Label Filter on the Prs page and on the issues page is always populated with all the Prs or issues' labels. Also if we select one filter, the dropdown always shows all the options.

Fixes: #1341


hdJerry (approved):
Nicely done sir

AllStackDev1 (approved):

Issue #1346, opened 2023-01-31
[solidstart-tanstack-tailwind] 02 Setup page structure


kenslachtajr:
# Setup page structure

## Background

It is important that before commencing work on creating application pages that we establish the structure of the pages for desktop and mobile grid

## Acceptance

- [ ] Set up structure for pages
  - [ ] Desktop
  - [ ] Mobile

## Notes

- Pages need to scale correctly in all views.

PR #1358, opened 2023-02-01
solidstart-feat: add responsive page layout

Megio:
Added a common responsive page layout.
For now, no responsive styles were added, but whenever we'll have to add some responsive changes to the layout, the only component that we have to change should be the Layout component.

Fixes: #1346


solidstart-tanstackquery-tailwind-modules/src/root.tsx
```diff
@@ -37,9 +38,11 @@ export default function Root() {
         <QueryClientProvider client={queryClient}>
           <Suspense>
             <ErrorBoundary>
-              <Routes>
-                <FileRoutes />
-              </Routes>
+              <Layout>
```
hdJerry, line 41: The layout will be a problem because it means signin page will have the header too. There has to be a way solid start handles layouts maybe

hdJerry: We can also choose to tackle this when the auth is merged, depending on you and we can do a little research and if it's taken too long we can merge and do a follow-up ticket. What do you think?

Megio: yes, I think this is the best thing to do. Merge the auth branch and then see how to manage the common layout. I took a look at solid start docs, but it seems there is no best practice to manage layout. They have a `<Outlet>` component that could be used but using the routing with `<A>` link makes it impossible I think. But we can definitely take another look at this after the auth branch is merged.

solidstart-tanstackquery-tailwind-modules/src/components/layout.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const Layout = (props: LayoutProps) => {
+  const { authStore }: { authStore: StoreProps } = useAuth();
+
+  return (
+    <>
+      <Show when={authStore.isAuthenticated}>
+        <Header />
+      </Show>
+      <main class="min-h-screen bg-gray-100">{props.children}</main>
```
hdJerry, line 18: https://www.solidjs.com/tutorial/props_children


Check

hdJerry (approved):
Thanks Sir

Issue #1351, opened 2023-02-01
[solidstart-tanstack-tailwind] 05 Get gists from Github API


kenslachtajr:
# Get Gists from GitHub API

## Background

We want to show the user a list of their gists.

## Acceptance

- [x] Get gists from GitHub API
- [x] For display on main page; data needed:
    - [x] Name of gist
    - [x] html_url

PR #1371, opened 2023-02-03
Solidstart feature: gists fetch

Megio:
Add query and fetch function for retrieve gists data

closes: #1351


AllStackDev1 (dismissed):
LGTM

solidjs-tailwind/src/services/get-gists.js
```diff
@@ -24,14 +24,14 @@ const getGists = async () => {
       (_acc, file) =>
```
SDaian, line 24: we shouldn't modify other Pod's file, please revert this change

SDaian (changes requested):

solidjs-tailwind/src/services/get-gists.js
hdJerry: This is a result of the husky not being correctly configured. But has been now. Kindly revert, please. THanks sir

solidstart-tanstackquery-tailwind-modules/src/services/api.ts
```diff
@@ -1,7 +1,7 @@
 export type ApiProps = {
   url: string;
   query: string | null;
-  variables: Record<string, string | number | null> | null;
+  variables?: Record<string, string | number | null> | null;
```
hdJerry, line 4: This will likely conflict with some of your recent changes

SDaian (approved):
LGTM

PR #1374, opened 2023-02-03
Solidstart/get organization repos api

AllStackDev1:
Closes: #1354 

# Get Organization repositories from GitHub API

## Background

There will be a view for an organization’s profile, which, similar to the user’s profile, will list all of the organization’s repositories.

## Acceptance

- [x] Get organization repositories from GitHub API
- [x] For display on organization page; data needed:
    - [x] Repo name
    - [x] description
    - [x] language
    - [x] star count
    - [x] fork count
    - [x] last updated
    - [x] visibility tag

## Notes

- This query should start sorted by most recently updated.

### DIFF
--- a/solidstart-tanstackquery-tailwind-modules/src/services/get-org-repos.ts
+++ b/solidstart-tanstackquery-tailwind-modules/src/services/get-org-repos.ts
@@ -1,0 +1,33 @@
+import FetchApi from './api';
+import { useAuth } from '../auth';
+import { GITHUB_GRAPHQL } from '../utils/constants';
+import { ORGANIZATION_REPOS_QUERY } from './queries/org-repos';
+import { OrgRepoInfo } from '~/types/org-repos';
+
+type Variables = Record<string, string | number | null> | null;
+type Response = {
+  data: OrgRepoInfo;
+};
+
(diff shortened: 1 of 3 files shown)


hdJerry (approved):

solidstart-tanstackquery-tailwind-modules/src/services/get-org-repos.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type Variables = Record<string, string | number | null> | null;
+type Response = {
+  data: OrgRepoInfo;
+};
+
+const getOrgRepos = async (variables: Variables) => {
+  const { authStore } = useAuth();
+
+  const data = {
```
hdJerry, line 15: Please add type to this data

hdJerry (changes requested):

solidstart-tanstackquery-tailwind-modules/src/services/get-org-repos.ts
AllStackDev1: We don't need to do this since the type has been declared in the useAuth(), it carried the types with it.

You can see the image below;

image

hdJerry (approved):

PR #1382, opened 2023-02-06
[solidstart-tanstack-tailwind] 13 Create gists panel

hdJerry:
# Create gists panel

## Background

As a user, I should see an unordered list of my gists on the main home page.

## Acceptance

- [x] Gist title
- [x] unordered list of user gists
- [x] clicking a gist opens that file in a new tab

## Result

Screenshot 2023-02-06 at 13 07 43

### DIFF
--- a/solidstart-tanstackquery-tailwind-modules/src/components/GistPanel/Gists.spec.tsx
+++ b/solidstart-tanstackquery-tailwind-modules/src/components/GistPanel/Gists.spec.tsx
@@ -5,7 +5,7 @@ import { GistsData } from './index';
 import { gists } from './data';
 
 describe('User gist card', () => {
-  let wrapper: any;
+  let wrapper;
 
   beforeEach(async () => {
     wrapper = await render(() => (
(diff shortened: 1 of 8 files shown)


solidstart-tanstackquery-tailwind-modules/src/components/GistPanel/Gists.spec.tsx
```diff
@@ -0,0 +1,26 @@
...
+describe('User gist card', () => {
+  let wrapper: any;
+
+  beforeEach(async () => {
+    wrapper = await render(() => (
+      <Router>
+        <GistsData gists={gists} />
+      </Router>
+    ));
+  });
```
Megio, line 16: If we move the assignation of wrapper inside every single tests instead of using beforeEach (we only have a few tests, so it's not a big effort I think); we can use typescript's type inference in order to have wrapper type and not use `any`.

hdJerry: I understand, but I think this is fine and it's a good practice and we don't need to do any render again.

Megio (approved):
LGTM

Megio (approved):
LGTM

PR #1408, opened 2023-02-08
Solidstart/create single repo view about section

AllStackDev1:
Close: #1389

image

# Create single repo view about section

## Background

A sub-component of the single repo page, this component focuses on the about section. It should show the user details about the repository, and any links made available to the user from that section.

## Acceptance

- [x] About title
- [x] repo description
- [x] link icon and url for deployed version (if available)
- [x] open book icon and readme title
- [x] divider line

### DIFF
--- a/solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/Description.tsx
+++ b/solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/Description.tsx
@@ -4,7 +4,10 @@ interface DescriptionProps {
 
 export const Description = (props: DescriptionProps) => {
   return (
-    <span data-testid="repo file explorer description" class={props.text ? 'italic' : undefined}>
+    <span
+      data-testid="repo file explorer description"
+      class={props.text ? 'italic' : undefined}
+    >
       {props.text || 'No description, website, or topics provided.'}
     </span>
   );
--- a/solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/Description.tsx
+++ b/solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/Description.tsx
@@ -1,0 +1,20 @@
+import { Match, Switch } from 'solid-js';
+
+interface DescriptionProps {
+  text: string;
+}
+
+export const Description = (props: DescriptionProps) => {
+  return (
+    <Switch>
+      <Match when={props.text}>
+        <span data-testid="repo file explorer description">{props.text}</span>
+      </Match>
+      <Match when={!props.text}>
+        <span class="italic" data-testid="repo file explorer description">
+          No description, website, or topics provided.
+        </span>
+      </Match>
+    </Switch>
+  );
+};
--- a/solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/Description.tsx
+++ b/solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/Description.tsx
@@ -1,20 +1,11 @@
-import { Match, Switch } from 'solid-js';
-
 interface DescriptionProps {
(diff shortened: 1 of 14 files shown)


solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/RepoAbout.tsx
```diff
@@ -0,0 +1,34 @@
+import { BookOpenIcon } from '../Icons';
+import { Description } from './Description';
+import { HomepageUrl } from './HomePageUrl';
+import { Topics } from './Topics';
+import styles from './RepoAbout.module.css';
+
+export const RepoAboutWidget = () => {
+  const info = {
```
hdJerry, line 8: there is a demo data in data.ts already

hdJerry (changes requested):

solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/Description.tsx
```diff
@@ -0,0 +1,20 @@
+import { Match, Switch } from 'solid-js';
+
+interface DescriptionProps {
+  text: string;
```
Megio, line 4: Inside the render, we check if this prop could be undefined, so probably here we must type is as `text?:string`

solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/Description.tsx
```diff
@@ -0,0 +1,20 @@
...
+    <Switch>
+      <Match when={props.text}>
+        <span data-testid="repo file explorer description">{props.text}</span>
+      </Match>
+      <Match when={!props.text}>
+        <span class="italic" data-testid="repo file explorer description">
+          No description, website, or topics provided.
+        </span>
+      </Match>
+    </Switch>
```
Megio, line 18: Don't know if here is a good practise to use `Switch` and `Match`. 
Probably we just need to do something like: 

```suggestion
   <span data-testid="repo file explorer description" class={props.text ? 'italic' : undefined}>
      {props.text || 'No description, website, or topics provided.'}
    </span>
```

AllStackDev1: I wanted to be more solid but this simliplify it.

solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/HomePageUrl.tsx
```diff
@@ -0,0 +1,21 @@
+import { A } from '@solidjs/router';
+import { Show } from 'solid-js';
+import { LinkIcon } from '../Icons';
+import styles from './RepoAbout.module.css';
+
+interface HomepageUrlProps {
+  homepageUrl: string;
```
Megio, line 7: Also here, if we put a condition on this prop we can say that this could be undefined so `homepageUrl?: string`

solidstart-tanstackquery-tailwind-modules/src/components/RepoAbout/Topics.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const Topics = (props: TopicsProps) => {
+  return (
+    <>
+      {props.topics.length === 0 ? null : (
+        <div class="space-y-1">
+          <For each={props.topics}>
+            {(topic) => <span class={styles.topic}>{topic}</span>}
+          </For>
+        </div>
+      )}
```
Megio, line 17: We can use `<Show when=` instead with a fallback as `null`

Megio (approved):
LGTM

PR #1412, opened 2023-02-09
Solidstart/type filter functionality

hdJerry:
# Implement type filter functionality

## Background

In another ticket, a component was created with the type dropdown button. The purpose of this ticket is to add the functionality and populate the fields for that button. The options that should be available to the user are listed below - when a user selects an option, the repository list on the page should be filtered to only include ones that match the user’s selection.

## Acceptance

- [x] only one selection allowed, noted with a checkmark icon

## result

Screenshot 2023-02-10 at 12 19 24

Screenshot 2023-02-10 at 12 34 39


solidstart-tanstackquery-tailwind-modules/src/components/RepoFilter/RepoFilter.tsx
```diff
@@ -24,21 +23,17 @@ const RepoFilter = (props: RepoFilterProps) => {
...
     { repoBtnText: 'New', languages: languageOptions, filteredRepoCount: 0 },
     props
   );
 
   const selectLanguage = () => null;
-  const selectType = () => null;
+  const selectType = (value: string) => setFilterType(value);
   const selectSort = () => null;
 
-  createEffect(() => {
```
hdJerry, line 38: This was removed because there is a ticket to bring all functionality together

Megio (approved):
LGTM

PR #1416, opened 2023-02-09
Solidstart-feat: search functionality

Megio:
Added repos search functionality in profile page and in org page

Screenshot 2023-02-10 at 11 13 58

Screenshot 2023-02-10 at 11 13 46

Fixes: #1378

### DIFF
--- a/solidstart-tanstackquery-tailwind-modules/src/utils/useRepoSortFilter.ts
+++ b/solidstart-tanstackquery-tailwind-modules/src/utils/useRepoSortFilter.ts
@@ -1,7 +1,5 @@
 import { UserRepo } from '~/types/user-repo-type';
-import {
-  search,
-} from '../components/RepoFilter/RepoFilter.store';
+import { search } from '../components/RepoFilter/RepoFilter.store';
 import { repoDataFilteredBySearch } from './searchFunction';
 
 const useRepoSortFilter = (repos: UserRepo[]) => {
--- a/solidstart-tanstackquery-tailwind-modules/src/utils/useRepoSortFilter.ts
+++ b/solidstart-tanstackquery-tailwind-modules/src/utils/useRepoSortFilter.ts
@@ -1,0 +1,15 @@
+import { UserRepo } from '~/types/user-repo-type';
+import {
+  search,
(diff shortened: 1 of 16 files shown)


hdJerry (changes requested):
The organisation page is broken, where you able to test the search filter there also?

solidstart-tanstackquery-tailwind-modules/src/utils/useRepoSortFilter.ts
```diff
@@ -0,0 +1,13 @@
+import { UserRepo } from '~/types/user-repo-type';
+import { search } from '../components/RepoFilter/RepoFilter.store';
+import { repoDataFilteredBySearch } from './searchFunction';
+
+const useRepoSortFilter = (repos: UserRepo[]) => {
```
hdJerry, line 5: not sure why we need this file as we can compress it into the repoDatafilterbyserach function

hdJerry (changes requested):

solidstart-tanstackquery-tailwind-modules/src/utils/useRepoSortFilter.ts
Megio: yes, I add this only because here we can manage also all the other filter and sort functions we have and return the right result according to the filter or sort or search parameters. Is redundant for the search only, but it could be useful when we add the other filtering and sorting functionalities

hdJerry: it's like a circle because repoDataFilteredBySearch is imported in useRepoSortFilter and useRepoSortFilter is imported in the main file it's used

Megio: you are right, I followed the other kit's organization, but we can define our search sort and filter function directly inside the `useRepoSortFilter` instead of having 4 separate files...

hdJerry: Oh, i see where you are coming from. AHAHAHA

hdJerry (approved):

PR #1458, opened 2023-02-10
Fix/gists display

AllStackDev1:
Fix gists not displaying in the home page 

Resolves: #1461

# Description
- The user gists where not been displayed on the home page, always showing User does not have any gists but I had gists.

### DIFF
--- a/solidstart-tanstackquery-tailwind-modules/src/services/get-gists.tsx
+++ b/solidstart-tanstackquery-tailwind-modules/src/services/get-gists.tsx
@@ -7,7 +7,11 @@ import { Gists } from '~/types/gists-type';
 
 interface Response {
   data: {
-    gists: Gists[];
+    viewer: {
+      gists: {
+        nodes: Gists[];
+      };
+    };
   };
 }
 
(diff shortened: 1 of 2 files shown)


solidstart-tanstackquery-tailwind-modules/src/services/get-gists.tsx
```diff
@@ -22,7 +26,31 @@ const getGists = async () => {
...
+                description: gist.description,
+                name: file.name || gist.name,
+                url: gist.url,
+              },
+            ]
+          : acc,
+      []
+    );
+    return [...acc, ...gists];
+  }, []);
```
Megio, line 51: We can type this function as:

```suggestion
  const gists = resp.data?.viewer.gists.nodes?.reduce((acc: Gists[] | { name: string }[], gist) => {
    if (!gist) {
      return acc;
    }
    const files = gist.files ?? [];
    const gists = files.reduce(
      (_acc: { name: string }[], file) =>
        file
          ? [
            ..._acc,
            {
              id: gist.id,
              description: gist.description,
              name: file.name || gist.name,
              url: gist.url,
            },
          ]
          : acc,
      []
    );
    return [...acc, ...gists];
  }, []);
```

Megio (approved):
LGTM

PR #1483, opened 2023-02-16
feat: adds readme content

### LINKED ISSUE #898: [Showcases API] Return README Content In A Usable Form
# Acceptance Criteria

- Readme: /repos/${owner}/${repoName}/readme
- The `README` will need some massaging to properly read the data. Using the default call listed will give you an encoded version of the content. You can pass a header to get the raw data or an HTML version.

WillHutt:
Resolves #898 
- adds readme
- gets formatters working and tweaks typedefs

image


jdwilkin4 (commented):
I tested this out locally with a few of the starter showcases and I think it looks much better.

The only issues I was running into were with images.
Is it possible to have a set width and height for images in the readme?

Sometimes, some of the repos would have images spill out of the container. 
Screenshot 2023-02-16 at 4 57 43 PM

I don't  know if that should be handled in this PR or maybe opening up a separate issue for another PR.

WillHutt:
Hmm, I think that would be a different ticket to resolve the front-end issue. 

I'm not sure if there is a way to set the width and height for the images inside the README, but I wouldn't be surprised. 

I believe the current issue could be due to the containers not setting a specific width and height themselves, so the images don't respect their container size.

jdwilkin4 (approved):
Looks good 👍

Issue #1475, opened 2023-02-15
[solidstart-tanstack-tailwind] Followers and following counts not displaying correct values


kenslachtajr:
# Description
- Followers and Following counts not displaying correct values - always zero

![Image](https://user-images.githubusercontent.com/1828602/219149228-e3841b73-1477-4c69-8085-4ce079dfe06e.png)

PR #1486, opened 2023-02-16
Solidstart/followers following count

AllStackDev1:
Closes: #1475

image

# Description
- Followers and Following counts not displaying correct values - always zero

Image

### DIFF
--- a/solidstart-tanstackquery-tailwind-modules/src/components/UserProfile/index.tsx
+++ b/solidstart-tanstackquery-tailwind-modules/src/components/UserProfile/index.tsx
@@ -9,103 +9,100 @@ import { Icon } from 'solid-heroicons';
 import { TwitterIcon } from '../Icons';
 import OrgList from './OrgList';
 import styles from './UserProfile.module.css';
-import { Show } from 'solid-js';
+import { Show, splitProps } from 'solid-js';
+import { UserProfileProps } from '~/types/user-profile-type';
 
-export interface UserProfileProps {
-  avatarUrl: string;
-  name: string;
-  username: string;
-  bio: string;
-  followers: number;
-  following: number;
-  starredRepos: number;
-  company: string;
-  location: string;
-  websiteUrl: string;
-  twitterUsername: string;
-  organizations: {
-    nodes: {
-      avatarUrl: string;
-      name: string;
-      login: string;
-    }[];
-  };
-}
+const UserProfile = (props: UserProfileProps) => {
+  const [local] = splitProps(props, [
+    'avatarUrl',
+    'name',
+    'username',
+    'bio',
+    'followers',
+    'following',
+    'starredRepositories',
+    'company',
+    'location',
+    'websiteUrl',
+    'twitterUsername',
+    'organizations',
+  ]);
 
-const UserProfile = (userProfileProps: UserProfileProps) => {
   return (
     <div>
       <img
-        src={userProfileProps.avatarUrl}
+        src={local.avatarUrl}
         alt="Avatar"
         width={260}
         height={260}
         class={styles.avatar}
       />
       <h1 class="mt-2">
-        <div class={styles.name}>{userProfileProps.name}</div>
-        <div class={styles.username}>{userProfileProps.username}</div>
+        <div class={styles.name}>{local.name}</div>
+        <div class={styles.username}>{local.username}</div>
       </h1>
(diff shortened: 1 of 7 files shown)


solidstart-tanstackquery-tailwind-modules/src/components/UserProfile/index.tsx
```diff
@@ -9,103 +9,100 @@ import { Icon } from 'solid-heroicons';
 import { TwitterIcon } from '../Icons';
 import OrgList from './OrgList';
 import styles from './UserProfile.module.css';
-import { Show } from 'solid-js';
+import { Show, splitProps } from 'solid-js';
+import { UserProfileProps } from '~/types/user-profile-type';
```
BigAB, line 13: Why would you move the props for this component to a separate file?

I think the props type should be colocated  with the component.

solidstart-tanstackquery-tailwind-modules/src/components/UserProfile/index.tsx
```diff
@@ -9,103 +9,100 @@ import { Icon } from 'solid-heroicons';
...
-  organizations: {
-    nodes: {
-      avatarUrl: string;
-      name: string;
-      login: string;
-    }[];
-  };
-}
+const UserProfile = (props: UserProfileProps) => {
+  const [local] = splitProps(props, [
```
BigAB, line 16: Is splitting here needed? It looks like you take ever property in the `local`

solidstart-tanstackquery-tailwind-modules/src/services/get-user-profile.ts
```diff
@@ -2,15 +2,15 @@ import FetchApi, { ApiProps } from './api';
...
+import { UserProfileProps } from '~/types/user-profile-type';
 
 type UserProfileVariables = {
   username: string;
 };
 
 type Response = {
   data: {
-    user: UserProfile;
+    user: UserProfileProps;
```
BigAB, line 13: So in general, the props of a component, or params of a function call, would (or can) be a different type than the entity type itself.

The `user` should probably be either type `User` or maybe `Profile` or `UserProfile` but it being `UserProfileProps` doesn't seem right.

If you had a `UserProfile` type, and the `Profile` components PropsType happened to exactly match the `UserProfile` type, than I would possible just have the Props be type as a `UserProfile` but the other way around seems backwards

solidstart-tanstackquery-tailwind-modules/src/types/user-profile-type.ts
```diff
@@ -1,21 +1,21 @@
-export interface UserProfile {
+export interface UserProfileProps {
```
BigAB, line 1: As noted above, this change does not seem like a good idea.

I would keep the component props type separate, and maybe even be a subset or extension of this `UserProfile` type, but don't put the props type here

BigAB (changes requested):
These comments are all pretty much related to the same thing.

solidstart-tanstackquery-tailwind-modules/src/services/get-user-profile.ts
AllStackDev1: Sure will update it.

solidstart-tanstackquery-tailwind-modules/src/components/UserProfile/index.tsx
AllStackDev1: Not really. updating it.

AllStackDev1: Splitting actually isn't needed here.

solidstart-tanstackquery-tailwind-modules/src/components/UserProfile/index.tsx
AllStackDev1: The props is been used in other places.

solidstart-tanstackquery-tailwind-modules/src/types/user-profile-type.ts
AllStackDev1: So this type also matches the API result we get from the query, and fits precisely with the props the component requires, that this why it is declared in the types/ to be used in places it is needed.

BigAB (approved):
I think this is good

PR #1487, opened 2023-02-16
update user profile card to show missing data points

AllStackDev1:
Closes: #

image

# Description
- Followers and Following counts not displaying correct values - always zero

Image


BigAB (approved):
LGTM

PR #1488, opened 2023-02-16
Solidstart/missing components

AllStackDev1:
Closes: #1473

image

image

# Description
- Missing tab header and pagination on the profile page

Image

Image

### DIFF
--- a/solidstart-tanstackquery-tailwind-modules/src/components/ProfileNav/ProfileNav.spec.tsx
+++ b/solidstart-tanstackquery-tailwind-modules/src/components/ProfileNav/ProfileNav.spec.tsx
@@ -3,7 +3,7 @@ import { render } from 'solid-testing-library';
 import { describe, expect, it } from 'vitest';
 import ProfileNav from './ProfileNav';
 
-describe('RepoCard for profilepage', () => {
+describe('Profile Navigation', () => {
   it('should mount', async () => {
     const wrapper = await render(() => (
       <Router>
@@ -11,5 +11,15 @@ describe('RepoCard for profilepage', () => {
       </Router>
     ));
     expect(wrapper).toBeTruthy();
+  });
+
+  it('should find menu item text', async () => {
+    const wrapper = await render(() => (
+      <Router>
+        <ProfileNav class="border-none" pathname={'/'} />
+      </Router>
+    ));
+    const text = await wrapper.getByText('Repositories');
+    expect(text).toBeDefined();
   });
 });
--- a/solidstart-tanstackquery-tailwind-modules/src/components/ProfileNav/ProfileNav.spec.tsx
+++ b/solidstart-tanstackquery-tailwind-modules/src/components/ProfileNav/ProfileNav.spec.tsx
@@ -1,0 +1,16 @@
+import { Router } from '@solidjs/router';
+import { render } from 'solid-testing-library';
+import { describe, expect, it } from 'vitest';
+import ProfileNav from './ProfileNav';
+
+describe('RepoCard for profilepage', () => {
+
+  it('should mount', async () => {
+    const wrapper = await render(() => (
+      <Router>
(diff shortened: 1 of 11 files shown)


solidstart-tanstackquery-tailwind-modules/src/components/ProfileNav/ProfileNav.spec.tsx
```diff
@@ -0,0 +1,15 @@
...
+import ProfileNav from './ProfileNav';
+
+describe('RepoCard for profilepage', () => {
+  it('should mount', async () => {
+    const wrapper = await render(() => (
+      <Router>
+        <ProfileNav class="border-none" pathname={'/'} />
+      </Router>
+    ));
+    expect(wrapper).toBeTruthy();
```
BigAB, line 13: I am not sure there is value in tests like this.

I don't think it tests anything, and I am not sure by the by the `describe` string what it is supposed to be testing

solidstart-tanstackquery-tailwind-modules/src/components/UserRepos/UserRepos.tsx
```diff
@@ -1,10 +1,16 @@
 import { For, Show, splitProps } from 'solid-js';
 import { RepoCard } from '~/components/RepoCard';
 import { RepoFilter } from '~/components/RepoFilter';
+import { Pagination } from '../Pagination';
 import { RepoInfos } from '../ProfilePage/ProfilePage';
 
 const UserRepos = (props: RepoInfos) => {
-  const [local] = splitProps(props, ['repos', 'pageInfo', 'languages']);
+  const [local] = splitProps(props, [
```
BigAB, line 8: I actually don't understand why we use `splitProps` here at all, is there something I am missing?

solidstart-tanstackquery-tailwind-modules/src/routes/[login].tsx
```diff
@@ -51,7 +51,7 @@ const Profile = () => {
   );
 
   const queryRepos = createQuery(
-    () => ['query-repos'],
+    () => [`query-repos-${location.query?.after}_${location.query?.before}`],
```
Comment (BigAB) [inline solidstart-tanstackquery-tailwind-modules/src/routes/[login].tsx:54]: This is interesting, why did you choose to do it this way?

solidstart-tanstackquery-tailwind-modules/src/components/UserRepos/UserRepos.tsx
AllStackDev1: So solidjs provide a safe way to pick/destruct some props value while keeping its reactivity, but most times we don't need it because we are not destructing props.

AllStackDev1: But it is also not needed here

solidstart-tanstackquery-tailwind-modules/src/routes/[login].tsx
Comment (AllStackDev1) [reply solidstart-tanstackquery-tailwind-modules/src/routes/[login].tsx:54]: So this serves as a query key for TanStack, when the key changes it fires another API call.

Comment (BigAB) [reply solidstart-tanstackquery-tailwind-modules/src/routes/[login].tsx:54]: I guess what I meant was, was there a reason to do this as a string interpolation rather than something like:
```ts
createQuery(
    () => ['query-repos', { after: location.query?.after, before: location.query?.before }],
```

Comment (AllStackDev1) [reply solidstart-tanstackquery-tailwind-modules/src/routes/[login].tsx:54]: No apparent reasons, this is also valid.

solidstart-tanstackquery-tailwind-modules/src/components/ProfileNav/ProfileNav.spec.tsx
AllStackDev1: Test will be updated and more test written

vyktoremario (approved):

PR #1505, opened 2023-02-23
solidjs-fix: remove husky

Megio:
Remove Husky from solid js showcase

Fixes: #1502


danecando:
Looks like there's a failing unit test. Maybe GH actions needs an env for a mock api url?

solidjs-tailwind/msw/server.js
```diff
@@ -1,10 +1,8 @@
 import { setupServer } from 'msw/node';
 import { beforeAll, afterAll, afterEach } from 'vitest';
-import { mockReadme } from './mockReadMe';
 import { mockRepoInfo } from './mockRepoInfo';
-import { mockRepoTree } from './mockRepoTree';
 
-const handlers = [mockRepoInfo, mockReadme, mockRepoTree];
```
hdJerry, line 7: why was the mockReadme and mockRepoTree removed?

Megio: I saw that we use msw server only for one test (RepoAbout) and it was exactly the test that fails in the CI step. I saw that probably we had some issues with the mocked API call so I decided to use a different approach: I mocked context instead of API calls and now everything seems fine. Probably all msw folder is useless now, so I can remove it and then push again.

Megio: ok I saw that we use the mockRepo thing also for storybook, so I think that the only thing we don't need anymore is the msw server setup file

hdJerry (approved):

PR #1539, opened 2023-02-28
feat: Expo Auth sessions

markshenouda:


BigAB (commented):
I couldn't get this working, not sure what the redirect_url needed to be, and had type errors about missing types for `@env` module

PR #1553, opened 2023-03-08
Expo 1535 profile layout

AllStackDev1:
Closes: #1535 
Mobile
image

Web
image

# Create profile page layout

## Background

As a user, I can navigate to my profile page and see my profile information, as well as a list of all my repositories. I can also see options to search my repositories, sort them, and filter them. (Functionality does not need to work yet for these options, they just need to be visible.)

## Acceptance

- [x] 2 column layout
- [ ] sub-header spanning both columns, showing book icon and repositories title - should be sticky on scroll
- [x] Left side
    - [x] user profile card
- [x] Right side - 
    - [x] search / filter / sort component
    - [x] card for all user repositories


expo-zustand-styled-components/src/components/FilterDropdown/FilterDropdown.tsx
```diff
@@ -49,19 +49,16 @@ const FilterDropdown = ({
...
-              {selected === item ? (
-                <CorrectIcon color={colors.gray600} />
-              ) : (
-                <View style={{ marginRight: 16 }} />
-              )}
-              <Text>{item}</Text>
-            </DropdownOption>
-          )}
-        />
+        {items.map((item) => (
```
vyktoremario, line 52: I am just curious (And maybe to learn also) but is there a reason why you decide not to use the `<FlatList />` component?

expo-zustand-styled-components/src/navigation/index.tsx
```diff
@@ -41,7 +41,7 @@ function RootNavigator() {
 
   return (
     <>
-      <StatusBar style="auto" />
+      <StatusBar style="light" />
```
vyktoremario, line 44: Any reason why this is changed?

expo-zustand-styled-components/src/components/FilterDropdown/FilterDropdown.tsx
AllStackDev1: I think I added a comment in one of the places I used map() instead of FlatList. FLatlist is mostly recommended when you have huge data and you want to lazy load them into the UI when the user scrolls. This isn't needed here as the dropdown can only have a limited number of data. Also, another issue with FlatList is, using it inside of a scroll view.

vyktoremario (commented):
Left some comments/Questions

expo-zustand-styled-components/src/navigation/index.tsx
AllStackDev1: The header has a dark background color. this change affect the statusbar color.

vyktoremario (approved):

expo-zustand-styled-components/src/hooks/stores/useAuthStore.ts
```diff
@@ -10,20 +10,25 @@ interface IAuthStore {
...
+  logout: () => null,
+  toggleMenu: () => null,
 };
 
 const useAuthStore = create(
   persist<IAuthStore>(
-    (set) => ({
+    (set, get) => ({
       ...initialState,
+      toggleMenu: (v) => set(() => ({ isMenuOpen: v ?? !get().isMenuOpen })),
```
Megio, line 31: Is there any reason why we decided to put this function inside the `AuthStore` instead of leaving it in the component?

Megio (approved):

PR #1601, opened 2023-03-16
using demo data as single repo page doesn't exist yet

hdJerry:
# Create single repo issues tab view

## Background

As a user, if I click into the issues tab for a repository, I can see a list of cards containing information about issues for this repo. By default I land on the view for open issues, and can click on the words “open” or “closed” to change if I see currently open or closed issues. 

## Acceptance

- [x] section header shows:
    - [x] {number} open with icon
    - [x] {number} closed with icon
    - [x] label dropdown (does not need options)
    - [x] milestones dropdown (does not need options)
    - [x] sort dropdown (does not need options)
- [x] body shows either open or closed (depending on which is selected) issue cards
- [x] pagination navigation - < prev next > (does not need to function yet)

## Notes

Screenshot 2023-03-16 at 10 22 56


AllStackDev1 (dismissed):

hdJerry:
> LGTM but why not we implement the feedback from Dustin here too. i mean the label should be colour coded

I want them to be a separate ticket. This ticket took too long. I will let Ken know about it.

Issue #1620, opened 2023-03-17
[solidstart-tanstack-tailwind] Update README to include known issues with Storybook implementation


kenslachtajr:
# Description
There are known issues with the way Storybook integrates with SolidStart framework. As such, most of what is in Storybook are the basic reusable components. This should be explained so that consumers are aware of these limitations with the integration.

PR #1628, opened 2023-03-22
update readme for storybook issue

AllStackDev1:
Closes: #1620
# Description
There are known issues with the way Storybook integrates with SolidStart framework. As such, most of what is in Storybook are the basic reusable components. This should be explained so that consumers are aware of these limitations with the integration.


solidstart-tanstackquery-tailwind-modules/README.md
```diff
@@ -33,6 +33,11 @@ SolidStart is a meta framework for SolidJS that is built on top of the React API
 - [ESLint](https://eslint.org/) - Code linting
 - [Prettier](https://prettier.io/) - Code formatting
 
+#### Note*
+We encountered an issue using Storybook in this showcase involving routes, you can follow the discord discussion [here](https://discord.com/channels/722131463138705510/1072093687892426843)
```
vyktoremario, line 37: I think it makes sense to give a brief summary of what the issue is before providing a link to follow the discussion.

vyktoremario (changes requested):
Left a comment

solidstart-tanstackquery-tailwind-modules/README.md
AllStackDev1: Ok

kenslachtajr (dismissed):

Megio (commented):
I see also a check annotation in a file unrelated to this PR, but probably we can fix that here too:

in `dynamicColor.ts` we should remove that `any` in the `getRGB` function.
we can typed that parameter like this:

```
function getRGB(c: string | number) {
  return typeof c === 'string' ? parseInt(c, 16) : c;
}
```

AllStackDev1:
> I see also a check annotation in a file unrelated to this PR, but probably we can fix that here too:
> 
> in `dynamicColor.ts` we should remove that `any` in the `getRGB` function. we can typed that parameter like this:
> 
> ```
> function getRGB(c: string | number) {
>   return typeof c === 'string' ? parseInt(c, 16) : c;
> }
> ```

This causes another set of errors.

image

Not sure of the purpose of that function so won't want to touch it much. lol

solidstart-tanstackquery-tailwind-modules/README.md
```diff
@@ -33,6 +33,12 @@ SolidStart is a meta framework for SolidJS that is built on top of the React API
 - [ESLint](https://eslint.org/) - Code linting
 - [Prettier](https://prettier.io/) - Code formatting
 
+#### Note*
+We encountered an issue using Storybook in this showcase involving routes `Route layout not define`.
```
honzikec, line 37:
```suggestion
We encountered an issue using Storybook in this showcase involving routes `Route layout not defined`.
```

honzikec (changes requested):
I think I found a small typo.

hdJerry:
```
{
    "owner": "thisdot",
    "name": "starter.dev-github-showcases",
    "branch": "main",
    "path/": "angular-apollo-tailwind/projects"
}
```

not sure why Solid Start params object have path with a "/"  suddenly.

Megio (changes requested):
Nested folder navigation is broken, if you try to navigate inside a folder and then again inside another child folder, the second click doesn't work

Megio (approved):

PR #1633, opened 2023-03-23
File Viewer

AllStackDev1:
Close: #1579 

Mobile

Simulator Screen Recording - iPhone 14 - 2023-03-22 at 13 28 08

Web
ezgif-2-17f7854569

# Create file view

## Background

As a user, if I’m viewing a single repository and click on an individual file, I am navigated to a single page view of that file. This should look similar to how a file would look on GitHub itself.

## Acceptance

- [x] clicking on a file in a repo displays that file


expo-zustand-styled-components/src/services/get-repo-file.ts
```diff
@@ -16,12 +16,19 @@ type Response = {
...
-  } catch (err) {
-    useAppStore.setState({ isLoading: false, error: err.message });
+  const key = `${variables.name}-${variables.owner}-${variables.expression}`;
+  if (useRepoInfoStore.getState()._file.has(key)) {
+    const data = useRepoInfoStore.getState()._file.get(key);
+    useRepoInfoStore.setState({ isLoading: false, file: data });
+  } else {
+    try {
+      useRepoInfoStore.setState({ isLoading: true });
+      const resp = (await FetchApi({ query: REPO_FILE_QUERY, variables })) as Response;
```
vyktoremario, line 26: Just a suggestion and maybe for another PR. I feel it will make a lot of sense for FetchAPI to take a take generic type of the data it is meant to return. 
So an implementation like 
```ts
export function FetchApi<T>(): T {
// ....
}
```
And then it is easier to just pass the type here as 
```ts
await FetchApi<RepoFile>({ query: REPO_FILE_QUERY, variables })
```
Using the `as` keyword for casting types can easily introduce bugs

vyktoremario (approved):
LGTM

I just left a suggestion (not for this PR)

expo-zustand-styled-components/src/services/get-repo-file.ts
AllStackDev1: True, case of copying from other showcases

PR #1641, opened 2023-03-24
Web linking fix

AllStackDev1:
Closes: #1639
Closes: #1640 

- [x] Research and implementation for URL Integration/Linking for web #1639
- [x] Passing data via the route path instead of the store #1640

Summary
We can now navigate using the `<LinkButton to='' />`, which accepts a route link, you can check the `linkingConfiguration.ts` file to add more route linking after adding a new screen to the stack navigator.

- [x] All previous navigations are working as they should.
- [x] Data is been fetched properly
- [x] Reloading the app on the web doesn't go to the index page.

image

image


expo-zustand-styled-components/src/components/FileViewer/FileViewer.tsx
```diff
@@ -12,8 +12,8 @@ import FileText from './FileText';
 
 import { ByteSize, Containter, Header, LineCount } from './FileViewer.styles';
 
-const FileViewer = () => {
-  const { path, owner, name, file, branch } = useRepoInfoStore();
+const FileViewer = ({ path, branch }: { path: string; branch: string }) => {
+  const { file, name, owner } = useRepoInfoStore();
```
hdJerry, line 16: i belive the owner and name should be gotten from the URL params if we are really moving the stored values to the url

expo-zustand-styled-components/src/components/FileTree/FileTree.tsx
```diff
@@ -1,54 +1,55 @@
...
 
 import { Cell, LinkText, Containter } from './FileTree.styles';
 import { colors } from '../../utils/style-variables';
 import { FolderIcon, DocumentIcon } from '../Icons';
+import LinkButton from '../LinkButton/LinkButton';
 
-const FileTree = () => {
-  const { path, owner, name, tree, branch } = useRepoInfoStore();
+const FileTree = (props: { path?: string; branch?: string }) => {
+  const { owner, name, tree, branch } = useRepoInfoStore();
```
hdJerry, line 13: i belive the owner and name should be gotten from the URL params if we are really moving the stored values to the url

expo-zustand-styled-components/src/components/BranchNavigation/BranchNavigation.tsx
```diff
@@ -14,38 +14,38 @@ import {
   ButtonTextStyled,
 } from './BranchNavigation.styles';
 import { useRepoInfoStore } from '../../hooks/stores';
+import LinkButton from '../LinkButton/LinkButton';
 
-const BranchNavigation = () => {
+const BranchNavigation = (props: { path?: string; branch?: string }) => {
   const { width } = useWindowDimensions();
-  const { path, name, branch } = useRepoInfoStore();
+  const { name, owner, branch } = useRepoInfoStore();
```
hdJerry, line 21: i belive the owner and name should be gotten from the URL params if we are really moving the stored values to the url

hdJerry (changes requested):
I just have one concern.
Nice work done.

expo-zustand-styled-components/src/components/FileViewer/FileViewer.tsx
AllStackDev1: image

From the image above the RepoNavigator has a path: ':owner/:name' and it has access to the name and owner path params, and we need to call getRepoInfo once. I noticed the subsequent screens don't have access to the owner and name path. not like the URL on the normal web, it is stored as JSON.

expo-zustand-styled-components/src/components/FileTree/FileTree.tsx
AllStackDev1: It isn't the same as the web normal url params. the owner and name were defined as the base URL, and only the base navigator(RepoNavigator) has access to them.

PR #1643, opened 2023-03-24
responsiveness completed

hdJerry:
# Description
- Currently, the repo sub-header is not responsive.

Image

Screenshot 2023-03-24 at 15 18 52

### DIFF
--- a/expo-zustand-styled-components/src/components/RepoActionButtons/RepoActionButtons.styles.ts
+++ b/expo-zustand-styled-components/src/components/RepoActionButtons/RepoActionButtons.styles.ts
@@ -7,11 +7,12 @@ type ScreenWidth = {
 };
 
 export const Container = styled.View<ScreenWidth>`
-  flex-wrap: wrap;
+  flex-wrap: nowrap;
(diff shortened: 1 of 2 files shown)


Megio (commented):
Everything looks good, the only thing I don't understand is why for this breakpoint we decided to go on a new line with the `public` tag. 

Screenshot 2023-03-24 at 17 13 29

Megio (approved):
LGTM

AllStackDev1 (changes requested):
The issue with this is when you run it on mobile, it doesn't work well. when you have a `name` that can fill the screen, it doesn't break into a new line. That piece of logic you remove handles that. 

image

hdJerry:
> The issue with this is when you run it on mobile, it doesn't work well. when you have a `name` that can fill the screen, it doesn't break into a new line. That piece of logic you remove handles that.
> 
> image
> 
> image As you can see.

The piece of logic was also breaking somethings also, 
There is a better way to handle it. I will look it and create a fall back ticket.

Thanks

PR #1690, opened 2023-03-28
pagination add to org profile and user profile

AllStackDev1:
Closes:  #1637

Added Pagination in orgs profile and user profile screen

The user menu dropdown component was updated as it hasn't been properly linked, resulting in an undefined path in the URL path.

https://user-images.githubusercontent.com/36667083/228393653-4e280e86-afe4-44ad-8d50-fb457965c04b.mov

https://user-images.githubusercontent.com/36667083/228393670-5635b5f6-fb3d-430c-85e3-160e171d6b71.mp4

### DIFF
--- a/expo-zustand-styled-components/src/components/Header/Header.tsx
+++ b/expo-zustand-styled-components/src/components/Header/Header.tsx
@@ -1,34 +1,22 @@
 import React from 'react';
-import { TouchableOpacity } from 'react-native';
-import { ParamListBase } from '@react-navigation/native';
-import { NativeStackNavigationProp } from '@react-navigation/native-stack';
 
 import { StyledHeader, StyledHeaderContainer } from './Header.styles';
 import UserDropdown from '../UserDropdown';
 import { GitHubLogo } from './GitHubLogo';
 
 import { useAuthStore } from '../../hooks/stores';
+import LinkButton from '../LinkButton/LinkButton';
 
-const Header = ({
-  width,
-  navigation,
-}: {
-  width: number;
-  navigation: NativeStackNavigationProp<ParamListBase, string, undefined>;
-}) => {
+const Header = ({ width }: { width: number }) => {
   const { toggleMenu } = useAuthStore();
 
   return (
     <StyledHeader screenWidth={width}>
       <StyledHeaderContainer screenWidth={width}>
-        <TouchableOpacity
-          onPress={() => {
-            toggleMenu(false);
-            navigation.navigate('Home');
-          }}>
+        <LinkButton to="/" onClick={() => toggleMenu(false)}>
           <GitHubLogo />
-        </TouchableOpacity>
-        <UserDropdown width={width} navigation={navigation} />
+        </LinkButton>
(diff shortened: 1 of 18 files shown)


hdJerry (changes requested):
When I click next in the pagination for the issue tab, the dropdowns disappear, and then I click next again it appears.

Also, it will be better to clear the before and after in the URL params when changing from open to closed issue or pr as the pagination/and API call works with that as well to avoid issues.


Screenshot 2023-03-28 at 15 24 44

expo-zustand-styled-components/src/components/Header/Header.tsx
```diff
@@ -1,34 +1,22 @@
...
+import LinkButton from '../LinkButton/LinkButton';
 
-const Header = ({
-  width,
-  navigation,
-}: {
-  width: number;
-  navigation: NativeStackNavigationProp<ParamListBase, string, undefined>;
-}) => {
+const Header = ({ width }: { width: number }) => {
```
hdJerry, line 10: The changes in this file have nothing to do with pagination, or does it?

expo-zustand-styled-components/src/components/IssuePullRequestCard/IssuePullRequestCard.tsx
```diff
@@ -10,32 +13,35 @@ import {
...
-import { colors } from '../../utils/style-variables';
 import PullRequestIcon from '../Icons/PullRequestIcon';
 import MergedPrIcon from '../Icons/MergedPrIcon';
 import ClosedPrIcon from '../Icons/ClosedPrIcon';
-import { Text, View } from 'react-native';
-import { format } from 'date-fns';
-import { Link } from '@react-navigation/native';
 import CommentIcon from '../Icons/CommentIcon';
+import IssuesIcon from '../Icons/IssuesIcon';
+
```
hdJerry, line 25: The changes in this file have nothing to do with pagination, why make the changes?

hdJerry (changes requested):
.

hdJerry (changes requested):
- In the profile/org repo page, when I click next and then prev, the pagination button is disabled.
- When I click next like 2 times and click prev, it goes back to the first page.

All these happened in both org and profile repo page

AllStackDev1:
> * In the profile/org repo page, when I click next and then prev, the pagination button is disabled.
> * When I click next like 2 times and click prev, it goes back to the first page.
> 
> All these happened in both org and profile repo page

All these are what we experience in other showcases, just that in this case we don't totally remove the pagination we simply disable them.

hdJerry:
> > * In the profile/org repo page, when I click next and then prev, the pagination button is disabled.
> > * When I click next like 2 times and click prev, it goes back to the first page.
> > 
> > All these happened in both org and profile repo page
> 
> All these are what we experience in other showcases, just that in this case we don't totally remove the pagination we simply disable them.

Thanks for letting us know sir.

hdJerry (approved):
Thanks.

PR #1694, opened 2023-03-28
fixing uri warning for image source

hdJerry:
# Description
- Need to ensure that the URL for whatever we are trying to display is available. If it is not available, there needs to be a placeholder. The property cannot accept an empty string or undefined.

### DIFF
--- a/expo-zustand-styled-components/src/components/Organization/About/About.styles.ts
+++ b/expo-zustand-styled-components/src/components/Organization/About/About.styles.ts
@@ -14,6 +14,7 @@ export const OrgLogoContainter = styled.View`
   border-width: 1px;
   border-radius: 4px;
   border-color: ${colors.gray200};
+  background-color: ${colors.gray300};
 `;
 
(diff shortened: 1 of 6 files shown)


AllStackDev1 (commented):
I would suggest we use a placeholder image/component because this might break the design. Consider the user avatar, which acts like a button, if the user doesn't have any image set, it will just be the dropdown arrow.

hdJerry:
> I would suggest we use a placeholder image/component because this might break the design. Consider the user avatar, which acts like a button, if the user doesn't have any image set, it will just be the dropdown arrow.

Can you be more specific and what do you mean by can break the design?

AllStackDev1:
> > I would suggest we use a placeholder image/component because this might break the design. Consider the user avatar, which acts like a button, if the user doesn't have any image set, it will just be the dropdown arrow.
> 
> Can you be more specific and what do you mean by can break the design?

image

It is cool.

AllStackDev1 (approved):

PR #1855, opened 2023-07-12
get top repos api

hdJerry:
## Background

Like the actual GitHub dashboard, we want to show a user their top repositories. The actual dashboard shows some other information, such as event feeds, but these aren’t available via the API easily so we opted for top repositories instead.

## Acceptance

- [x] Get top repositories from GitHub API
- [x] For display on main page; data needed:
    - [x] Repo name
    - [x] description
    - [x] language
    - [x] star count
    - [x] fork count
    - [x] last updated
    - [x] visibility tag (public / private)

## Notes

- If you’re using the GraphQL API, there’s a TopRepositories query you can run to retrieve this information easily. (example [GraphQL](angular-apollo-tailwind/src/app/gql/queries/user-top-repos.query.gql) )
- If you’re using the REST API, an equivalent query doesn’t exist. We’ve opted to use the following request instead. (example [rest API](angular-ngrx-scss/src/app/user/services/user.service.ts))

### DIFF
--- a/nuxt3-pinia-vuetify/hooks/useFetchApI.ts
+++ b/nuxt3-pinia-vuetify/hooks/useFetchApI.ts
@@ -1,0 +1,13 @@
+import { UseFetchOptions } from "nuxt/app";
+
+export default function useFetchAPI(url: string, opts?: UseFetchOptions<unknown>) {
+	const config = useRuntimeConfig();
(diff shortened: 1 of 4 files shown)


vyktoremario (commented):
Same suggestions as the other PR

vyktoremario (approved):

PR #1857, opened 2023-07-12
Feat profile api

hdJerry:
## Background

For the profile page, we want to fairly closely resemble the GitHub user page. This section is focused on the user “card”, where you can see profile information about the user.

## Acceptance

- [x] Get user info from GitHub API
- [x] For display on profile page; data needed:
    - [x] Photo
    - [x] Name (display name)
    - [x] username (login name)
    - [x] description
    - [x] followers
    - [x] following
    - [x] stars
    - [x] company info
    - [x] location info
    - [x] blog link
    - [x] organizations (need name and avatar)

### DIFF
--- a/nuxt3-pinia-vuetify/hooks/useFetchApI.ts
+++ b/nuxt3-pinia-vuetify/hooks/useFetchApI.ts
@@ -1,6 +1,6 @@
 import { UseFetchOptions } from "nuxt/app";
 
-export default function useFetchAPI(url: string, opts?: UseFetchOptions<unknown>) {
+export default function useFetchAPI(url: string, opts?: UseFetchOptions<Record<string, any>>): Record<string, any> {
 	const config = useRuntimeConfig();
 	const { GITHUB_API_URL } = config;
 	const token = useCookie('token');
--- a/nuxt3-pinia-vuetify/hooks/useFetchApI.ts
+++ b/nuxt3-pinia-vuetify/hooks/useFetchApI.ts
@@ -1,6 +1,6 @@
 import { UseFetchOptions } from "nuxt/app";
 
-export default function useFetchAPI(url: string, opts?: UseFetchOptions<Record<string, any>>): Record<string, any> {
+export default function useFetchAPI<T>(url: string, opts?: UseFetchOptions<Record<string, any>>): Record<string, any> | { data: T } {
 	const config = useRuntimeConfig();
 	const { GITHUB_API_URL } = config;
 	const token = useCookie('token');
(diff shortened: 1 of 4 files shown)


nuxt3-pinia-vuetify/store/profileStore.ts
```diff
@@ -0,0 +1,76 @@
...
+				this.user = {
+					...this.user,
+					orgs,
+				};
+			} catch (error: any) {
+				if (error && error?.response) {
+					throw error;
+				}
+
+				throw new Error('Error fetching user top repos');
```
vyktoremario, line 51: This too

nuxt3-pinia-vuetify/hooks/useFetchApI.ts
```diff
@@ -0,0 +1,13 @@
+import { UseFetchOptions } from "nuxt/app";
+
+export default function useFetchAPI(url: string, opts?: UseFetchOptions<unknown>) {
```
vyktoremario, line 3: This is just a suggestion, but what about specifying a generic return type for the hook?

```typescript
export default function useFetchAPI<DataT>(url: string, opts?: UseFetchOptions<DataT>) {
const config = useRuntimeConfig();
	const { GITHUB_API_URL } = config;
	const token = useCookie('token');
	const headers = {
		...(opts?.headers || {}),
		...(token && { Authorization: `Bearer ${token.value}` }),
	};

	return useFetch(url, { ...opts, headers, baseURL: GITHUB_API_URL });
}
```

So that you avoid casting the return type (using `as`)
```typescript
const { data } = await useFetchAPI<IUser>(url, {
					headers: {
						Accept: 'application/vnd.github+json',
					},
				});
```

vyktoremario (changes requested):
Left some comments and suggestion

nuxt3-pinia-vuetify/hooks/useFetchApI.ts
```diff
@@ -0,0 +1,13 @@
...
+	const config = useRuntimeConfig();
+	const { GITHUB_API_URL } = config;
+	const token = useCookie('token');
+	const headers = {
+		...(opts?.headers || {}),
+		...(token && { Authorization: `Bearer ${token.value}` }),
+	};
+
+	return useFetch(url, { ...opts, headers, baseURL: GITHUB_API_URL });
+}
```
vyktoremario, line 13:
```typescript
export default function useFetchAPI<T>(url: string, opts?: UseFetchOptions<Record<string, any>>): T {
	const config = useRuntimeConfig();
	const { GITHUB_API_URL } = config;
	const token = useCookie('token');
	const headers = {
		...(opts?.headers || {}),
		...(token && { Authorization: `Bearer ${token.value}` }),
	};

	return useFetch(url, { ...opts, headers, baseURL: GITHUB_API_URL });
}
```

vyktoremario: `T` is a generic. You can pass the type when you call the function.

vyktoremario (approved):

PR #1888, opened 2023-08-02
Angular ngrx scss/update angular

### LINKED ISSUE #1885: [angular-ngrx-scss]: Update Angular and other deps to latest version
### Is your feature request related to a problem? Please describe.

- The starter currently uses Angular 13.3. We should update it to use Angular 16 and update any other packages to their latest versions in the process.
- After we update the starter, we should update the showcase with the same deps.

The related issue: https://github.com/thisdot/starter.dev/issues/1321

### Describe the solution you'd like

Update dependencies to match the updated starter.Update dependencies to match the updated starter.

### Describe alternatives you've considered

Keeping deps out of date.

### Additional context

_No response_

honzikec:
## Summary of change

resolves #1885

- update Angular to 16
- fix issues that came form the update

## Checklist

- [x] These changes follow the [contributing guidelines](https://github.com/thisdot/starter.dev-github-showcases/blob/main/CONTRIBUTING.md)
- [x] I have verified this change works and introduces no additional issues to the best of my knowledge

Should be merged after [#1322 ](https://github.com/thisdot/starter.dev/pull/1322)


angular-ngrx-scss/package.json
```diff
@@ -18,55 +18,55 @@
...
+    "@angular/compiler": "16.1.7",
+    "@angular/core": "16.1.7",
+    "@angular/forms": "16.1.7",
+    "@angular/platform-browser": "16.1.7",
+    "@angular/platform-browser-dynamic": "16.1.7",
+    "@angular/router": "16.1.7",
+    "@ngrx/effects": "16.1.0",
+    "@ngrx/store": "16.1.0",
+    "@ngrx/store-devtools": "16.1.0",
     "@primer/octicons": "^16.3.1",
```
lindakatcodes, line 32: Can we go ahead and lock all the package versions in this file? Looks like most of the Angular ones are locked now, but we should be consistent and just lock everything to a specific version here.

angular-ngrx-scss/projects/prism/package.json
```diff
@@ -2,10 +2,10 @@
   "name": "prism",
   "version": "0.0.1",
   "peerDependencies": {
-    "@angular/common": "^13.0.0",
-    "@angular/core": "^13.0.0"
+    "@angular/common": "^16.1.0",
```
lindakatcodes, line 5: Can we lock these to the specific version as well? I always forget this package is in here and I don't know if there's reasons it may need to be a little more flexible. I think the main package.json being locked is more important than this one.

lindakatcodes (changes requested):
Overall seems to work great! Site starts locally and looks like it should, build completes, and tests pass. :) Nice job!

I left two small comments we should adjust if we can for the package.json files. And I have one question: do we want two different lock files here? There's a package-lock and yarn-lock. I don't particularly care which one stays but I feel like we should only have the one (my guess would be to keep the package-lock since a few of the scripts are set to use npm but if they still work fine with the yarn-lock I'm good with that).

Once we've got those little things resolved this should be good to go!

lindakatcodes (approved):
Looks great! 😁

PR #1894, opened 2023-08-08
chore: fix repo header links

hdJerry:
Repo header link is not working.

- route to the org page if the repo is an org repo when owner text is clicked
- route to the user profile page when owner text is clicked

Image


cra-rxjs-styled-components/src/components/sub-header/SubHeader.tsx
```diff
@@ -62,6 +62,10 @@ export default function SubHeader() {
 		return pathname.includes(matchPath);
 	};
 
+	console.log('====================================');
+	console.log(repo);
+	console.log('====================================');
+
```
Megio, line 68: console.log

PR #1914, opened 2023-08-14
Cra rxjs search functionality

hdJerry:
# Implement search functionality

## Background

In another ticket, a component was created with a search input field. The purpose of this ticket is to add the functionality to that input. A user should be able to write a query in the input, and the repositories listed on the page should be filtered based on that query after a short debounce time. If the user removes their query, either by clearing all the letters or clicking the X that should appear in the input, the repositories will be reset.

## Acceptance

- [x] Search should filter all queries for any that match the search query
- [x] search works automatically after a debounce time
- [x] small X should show at end of input when user has added text, that when clicked will remove the query

## Notes

### DIFF
--- a/cra-rxjs-styled-components/src/components/repo-filter/SearchInput.tsx
+++ b/cra-rxjs-styled-components/src/components/repo-filter/SearchInput.tsx
@@ -1,5 +1,21 @@
+import { useRepoFilter } from '../../context/RepoFilterContext';
 import { SearchTextInput } from './RepoFilter.styles';
 
 export default function SearchInput() {
-	return <SearchTextInput placeholder="Find a repository..." />;
+	const { search, setSearch } = useRepoFilter();
+
+	const handleChange = (e: any) => {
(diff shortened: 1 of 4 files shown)


cra-rxjs-styled-components/src/components/repo-filter/SearchInput.tsx
```diff
@@ -1,5 +1,21 @@
+import { useRepoFilter } from '../../context/RepoFilterContext';
 import { SearchTextInput } from './RepoFilter.styles';
 
 export default function SearchInput() {
-	return <SearchTextInput placeholder="Find a repository..." />;
+	const { search, setSearch } = useRepoFilter();
+
+	const handleChange = (e: any) => {
```
Megio, line 7:
```suggestion
const handleChange = (e: React.ChangeEvent<HTMLInputElement>)
```

Megio (approved):
LGTM

